PR #11, opened 2016-11-07
Babel and Rollup configuration.

bhough:
Closes https://github.com/styled-components/cynosure/issues/9


bhough:
@mxstbr copied the rollup configuration over from styled-components. Removed any dependencies on React for this build since it should not be needed. I also copied the babel.rc that was created, looks like it was addressing an issue in styled-components, let me know if you think that won't be needed for this.

.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    "latest"
+  ],
+  "plugins": [
+    "add-module-exports",
+    "transform-flow-strip-types"
```
mxstbr: Since we don't have flow types, we also don't need to strip them! Other than that, this looks great!

bhough: K. I also removed all the dependencies and rollup config code related to flow as well.

bhough:
@mxstbr gonna have to look at this a bit more in the morning. Fixed some issues with capitalization of file names in the repo causing things to work locally but not on Travis. But now running into an issue where Jest passes on both `src` and `lib` locally, but fails on `src` on Travis. Looks like it is not using es6 properly on Travis for some reason.

bhough:
@mxstbr nevermind, figured it out. Added babel-jest to the test suite, so now we can write our tests in ES6 as well. Odd that it would work locally without it, but could be to me having babel installed globally. This is ready to merge.

mxstbr:
Awesome, thanks!

PR #12, opened 2016-11-07
Rename project to polished

### LINKED ISSUE #10: Name
We need a better name.

While the meaning of the word `cynosure` is amazing, two hours after I first found the name I couldn't even remember it – not a good sign.

Any ideas?

`pedicure` isn't taken on npm yet, would fit with our whole 💅-thing and has something to do with styling, but it does have a slightly dirty connotation maybe? `import { clearfix } from 'pedicure'`

/cc @styled-components/design @styled-components/cynosure

mxstbr:
Closes #10 

I also published v0.0.0 of `polished` to avoid people stealing the name!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# `cynosure`
+# `polished`
 
 A lightweight toolset for writing styles in Javascript.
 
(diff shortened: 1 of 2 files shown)


bhough:
@mxstbr you will need to retrigger the build on this, the rename of the repo caused Travis to not be able to return the status of CI to github.

mxstbr:
Ah, good catch!

PR #15, opened 2016-11-09
Jest Snapshots and Doc Generation

bhough:
- Implements an example of snapshot testing with Jest.
- Implements automatic docs generation using JSDOCs .
- Adds watch task that both generates bundle and documentation.


bhough:
@mxstbr this is ready for review

.eslintrc
```diff
@@ -1,12 +1,24 @@
 {
-  "extends": "airbnb-base/legacy",
```
mxstbr: Why was this removed? We want those linting rules?

src/helpers/StripUnit.js
```diff
@@ -1,14 +1,26 @@
...
+import isInteger from '../internalHelpers/_isInteger'
+
 /**
- * Strips the unit from a given value.
- * @param {string} value - a string that is a number plus a unit of measure.
- * @return {number} - an integer without the unit.
+ * CSS to represent truncated text with an ellipsis.
+ * @static
+ * @param {string} [width = 100%] - Max-width for text to respect before being truncated.
+ * @return {Object} CSS as JS for truncating text.
```
mxstbr: These descriptions are all wrong I think? This should be "Strip unit of a given CSS value", "[value]" and it returns "Value without the unit"

mxstbr:
This is AMAZING! 🎉 Love it!

Two small questions, then we're good to go here!

.eslintrc
bhough: I actually removed this because it only covered ES5 standards. I'd like to replace it with the full base library (which covers ES6+) if you are fine with that.

src/helpers/StripUnit.js
bhough: Good catch. This is just an artifact of me testing the doc generator across multiple items.  I'll add proper documentation for this.

.eslintrc
bhough: We may want to add some overrides to their rules for our application as well. A couple examples:

- it throws errors for unnecessary quoted property names. Do we want to consistently quote them since there are so many hyphenated properties in css?

- It throws errors for missing spaces before function parentheses, but it looks like in `styled-components` we aren't enforcing that.

mxstbr: Yeah just use `eslint-config-airbnb`, the standard one.

Yeah the consistent quoted property names sounds like a good idea, and I don't like spaces before function parentheses so whichever way you want to go there is fine with me!

bhough:
alright @mxstbr take 2 :)

mxstbr:
Amazing work!

PR #16, opened 2016-11-14
Add wordWrap mixin

mxstbr:
Please review @styled-components/polished


src/mixins/wordWrap.js
```diff
@@ -0,0 +1,33 @@
...
+
+function wordWrap(wrap = 'break-word') {
+  const wordBreak = wrap === 'break-word' ? 'break-all' : wrap
+  return {
+    'overflow-wrap': wrap,
+    'word-wrap': wrap,
+    'word-break': wordBreak,
+  }
+}
+
```
bhough: Something I've been thinking about based on how Bourbon, Compass, etc.. are often used is how do we account for this traditional usage:

```scss
div {
  background: red;
  @include word-wrap;
}
```

Obviously this syntax doesn't work 1 to 1 in JS, but we should decide on how we want to tell users to handle this scenario. With this current implementation, we would have no way to in handle this.

src/mixins/test/wordWrap.test.js
```diff
@@ -0,0 +1,14 @@
...
+
+describe('wordWrap', function() {
+
+  it('should accept other values', function() {
+    expect(wordWrap('break-all')).toMatchSnapshot()
+  })
+
+  it('should default wrap to break-word', function() {
+    expect(wordWrap()).toMatchSnapshot()
+  })
```
bhough: Since we are going to be porting this over as "live" examples we should set up tests similar to this:

```javascript
expect({
  div: wordWrap()
}).toMatchSnapshot()
```

mxstbr: Ah sorry, good call!

I just realized that we should have two examples per util, demonstrating usage with styles as objects and `styled-components` independently:

```JS
// Styles as object usage
const styles = {
  ...wordWrap(),
}

// styled-components usage
const Main = styled.main`
  ${wordWrap()}
`
```

It would be cool if this was globally togglable (i.e. choose if the whole site displays `styled-components` or styles as objects usage) somewhere so we don't show both all the time. What do you think, is that feasible?

bhough: We should be able to do that, we have to do some customization to the docs template anyway before launch. 

Using spreads also solves my above comment about how to handle adding the rules to existing style blocks. I think it is more than safe to assume folks will be using babel with this anyway.

bhough: @mxstbr Let's change it now, so it is one less thing we will need to go back and change.

src/mixins/wordWrap.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Provides an easy way to change the `word-wrap` property
+ *
+ * @static
+ * @param {string} [wrap = 'break-word'] - Value for the word-break property
+ * @return {Object} CSS as JS.
+ * @example
+ * div {
+ *   wordWrap('break-all')
+ * }
```
bhough: I think this should be a spread?

mxstbr:
Done, ready to be reviewed again!

package.json
```diff
@@ -47,6 +47,7 @@
     "babel-loader": "^6.2.7",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-external-helpers": "^6.18.0",
+    "babel-plugin-transform-object-rest-spread": "^6.16.0",
```
mxstbr: Travis is complaining it can't find this babel plugin, but it's in the dependencies?

bhough:
@mxstbr looks like you need to add the `transform-object-rest-spread` to the `.babelrc` file so CI will pass.

.babelrc
```diff
@@ -3,6 +3,7 @@
     "latest"
   ],
   "plugins": [
+    "transform-object-rest-spread",
```
bhough: Odd, yeah it looks like it is seeing the reference, but not installing it properly?

bhough:
@mxstbr it looks like in the `.travis.yml` we aren't running `npm install` so it is not installing any of the new dependencies. I fixed that in: https://github.com/styled-components/polished/pull/17 and it resolved the issue.

mxstbr:
Okay let's merge #17 and #18 first, then rebase this on top of `master` and ship.

bhough:
This has been rebased, just needs review

mxstbr:
Awesome! You just need to submit an approved review, since I submitted the PR I can't do it 😉

bhough:
Is there a "duh" emoji. This is what I get for coding so early 😴

Issue #1, opened 2016-11-03
Initial list


mxstbr:
Below is a list of helpers we think would be useful to have in this toolset. A few notes before diving in and implementing them:

- We are aiming for 100% test coverage, so extensively test the utils you write
- Add helpful error messages and type checking based on the util @bhough will write in #4 
- These helpers should be method agnostic, i.e. not `styled-components` specific. This means they need to return css-in-js, by which I mean JavaScript objects, not CSS strings.

An example what `ellipsis` could return:

```JS
const Box = styled.div`
  ${ ellipsis('250px') }
`
```

```JS
{
  // Value from first argument
  width: '250px',
  whiteSpace: 'nowrap',
  overflow: 'hidden',
  textOverflow: 'ellipsis',
}
```

> Note that `styled-components` can handle simple styles as JavaScript objects perfectly fine, except you can't have pseudo-selectors and those sorts of things in the objects. We'll have to have a discussion what to do with the more complex of these helpers!

By returning objects, these also work for any CSS-in-JS method:

```JS
const styles = {
  ...ellipsis('250px'),
}
```

🎉

**Mixins**

- [x] Clearfix (@bhough, #22 )
- [x] Ellipsis (@bhough, #17)
- [x] Font Face (@bhough, #63)
- [x] Hide Text (@bhough, #19)
- [x] HiDPI Media Query (@bhough, #22)
- [x] Placeholder (@bhough, #42)
- [x] Radial Gradient (@bhough, #66)
- [x] Reset/Normalize (@bhough. #67)
- [x] Retina Image (@mxstbr, #50)
- [x] Selection (@bhough, #41)
- [x] Timing Functions (@bhough, #20)
- [x] Triangle (@machycek, #77)
- [x] Word Wrap (@mxstbr, #16)

**Shorthands**

- [x] Animation (@mxstbr, #29)
- [x] Background (@bhough, #64)
- [x] Background Image (@bhough, #64)
- [x] Border-Color (@bhough, #56)
- [x] Border-Radius (@bhough, #68)
- [x] Border-Style (@bhough, #56)
- [x] Border-Width (@bhough, #56)
- [x] Buttons (@bhough, #57)
- [x] Inputs (@bhough, #57)
- [x] Margin (@bhough, #56)
- [x] Padding (@bhough, #56)
- [x] Position (@bhough, #56)
- [x] Transition (@bhough, #65)
- [x] Size (@bhough, #43)

**Helpers**

[...]

PR #20, opened 2016-11-16
Add Timing Functions Mixin

bhough:
Addresses line item in #1


bhough:
@mxstbr this is ready for review, but wanted your thoughts on how we want to return static values like this. My thought is having a function that returns it would keep it consistent with similar mixins that do work behind the scenes (like strip unit). That way the API is always the same. Thoughts?

mxstbr:
I like it! What about adding a list of the functions we support to the docs?

bhough:
good idea!

bhough:
Added to docs and ready for review.

mxstbr:
This LGTM, except for this:

screen shot 2016-11-18 at 10 52 40

Any way to fix that overflowing list? I'm sure there could be a better display for them.

bhough:
So this turned out to be a bug of sorts with the theme instead of with how we are displaying it. Since we are going to need to customize the theme anyway, I brought the theme into the repo and fixed the bug on our copy that we will customize further later. 

Should be good to ship now once you approve.

PR #22, opened 2016-11-18
ClearFix and hiDPI Mixins

bhough:
Addresses line items in #1


bhough:
@mxstbr some new patterns here I want to make sure will fly with both styled-components and CSS as JS given some of the conversation around pseudo selectors for example.

mxstbr:
Sadly, neither of these examples will work with `styled-components`, we only support one level deep of styles-as-objects. 😕 Should we add support for that?

bhough:
Sorry been traveling. We should definitely considering adding support. Would be nice in general to have as much parody between what `styled-components` supports in strings and styles-as-objects.

Is there a technical limitation in `styled-components` that is keeping us from supporting pseudo selectors in styles-as-objects, or is it something that we can/should add?

mxstbr:
> Is there a technical limitation in styled-components that is keeping us from supporting pseudo selectors in styles-as-objects, or is it something that we can/should add?

It was a choice, we didn't want to encourage people to write their styles as objects. Ping @geelen, what do you think should we do here?

geelen:
Hmm, I'm not sure. Yes we can allow nested objects, that's no problem, it's just strange to render them to a string, then parse them back as PostCSS then unroll the nesting and inject the CSS. But I can't think of a better way so ¯\\\_(ツ)_/¯

mxstbr:
Will add support for that in core, then we can ship this.

mxstbr:
Added support for this in core, will be out soon, so let's get this rebased and land it! 💪

src/mixins/hiDPI.js
```diff
@@ -0,0 +1,41 @@
...
+ * @static
+ * @param {Number} [width = 1.3] - Device pixel ratio.
+ * @return {string} Represents the generated media query.
+ * @example
+ * // Styles as object usage
+ * PENDING....
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   ${hiDPI(1.5)}: {
```
mxstbr: This doesn't need the `:`, should just be:

```JS
const div = styled.div`
  ${hidDPI(1.5)} {
    width: 200px;
  }
`
```

bhough:
@mxstbr this should be ready to re-review.

src/mixins/clearFix.js
```diff
@@ -1,4 +1,40 @@
...
+/** @module Mixins */
+
+/**
+ * CSS to contain a float (credit to CSSMojo)
+ * @static
+ * @param {string} [parent = &] - Optional parent selector
+ * @return {Object} CSS as JS.
+ * @example
+ * // Styles as object usage
+ * PENDING...
```
mxstbr: Why is this pending? Should just be

```JS
const styles = {
  ...clearFix(),
}
```

src/mixins/clearFix.js
```diff
@@ -1,4 +1,40 @@
...
+ *   'clear': 'both',
+ *   'content': '',
+ *   'display': 'table'
+ * }
+ */
+
+function clearFix(parent: string = '&') {
+  const pseudoSelector = `${parent}::after`
+  return {
+    /* eslint-disable comma-dangle */
```
mxstbr: Why do we need to disable this? Just add a dangling comma?

bhough:
K @mxstbr third time's a charm.

mxstbr:
Awesome let's ship it!

PR #25, opened 2016-11-27
Add flowtype annotations

### LINKED ISSUE #23: Flow types
Planning on using flow type?

relekang:
Fixes #23


src/mixins/timingFunctions.js
```diff
@@ -1,38 +1,70 @@
...
+  'easeInOutCirc': 'cubic-bezier(0.785,  0.135, 0.150, 0.860)',
   'easeInOutCubic': 'cubic-bezier(0.645,  0.045, 0.355, 1.000)',
-  'easeInOutExpo':  'cubic-bezier(1.000,  0.000, 0.000, 1.000)',
-  'easeInOutQuad':  'cubic-bezier(0.455,  0.030, 0.515, 0.955)',
+  'easeInOutExpo': 'cubic-bezier(1.000,  0.000, 0.000, 1.000)',
+  'easeInOutQuad': 'cubic-bezier(0.455,  0.030, 0.515, 0.955)',
   'easeInOutQuart': 'cubic-bezier(0.770,  0.000, 0.175, 1.000)',
   'easeInOutQuint': 'cubic-bezier(0.860,  0.000, 0.070, 1.000)',
-  'easeInOutSine':  'cubic-bezier(0.445,  0.050, 0.550, 0.950)',
+  'easeInOutSine': 'cubic-bezier(0.445,  0.050, 0.550, 0.950)',
```
bhough: Small comment, but I think I prefer to have these all aligned so they are easier to compare the differences in the functions.

bhough:
@relekang Thanks for taking the time to do this. One small formatting comment. Also can you amend your commits all be chores. We are using `semantic-release`, so it will publish new versions based on those `feat` commits.

src/mixins/timingFunctions.js
relekang: This was fixed by eslint. I can change it back but would need to add `eslint-disable` comments for that rule in this file. Should I do that?

relekang:
Thanks for the review @bhough :v: I'll hopefully have time to fix things today or tomorrow.

> Also can you amend your commits all be chores.

Sure, but if we don't follow the [spec]() linked from the semantic-release docs we should probably have a contributing.md with which one of the commit types we allow.

[spec]: https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#

mxstbr:
> Sure, but if we don't follow the spec linked from the semantic-release docs we should probably have a contributing.md with which one of the commit types we allow.

We will totally follow them, but this isn't released yet and we don't want to push to npm just yet as there's nothing here, that's why :)

src/mixins/timingFunctions.js
bhough: If you could please.

relekang: The spacing is back now 🙂

relekang:
Thanks for the explanation. I updated all the commits to be chores 🙂

bhough:
Thanks for the contribution @relekang!

PR #29, opened 2016-12-08
chore(shorthands): Add animation shorthand

mxstbr:


bhough:
Looking at this again, we may not need this one. I think when bourbon initially implemented it the animation shorthand was lagging behind the individual properties. Question is if concatenating the array into a single string is worth a mixin.

mxstbr:
It's actually quite useful, consider this case:

```JS
const Rotating = styled.div`
  animation: ${constants.rotate} 1s ${constants.easing} 2s;
`
```

Already with two interpolations that's quite a pain to read and write. Compare with this case:

```JS
const Rotating = styled.div`
  ${animation(constants.rotate, '1s', constants.easing, '2s')}
`
```

Much nicer that! (even bigger difference in styles as objects) In fact, I'd say we have functions like these for all the common CSS functions – it's really quite useful if you have a lot of interpolated/shared variables like above!

bhough:
Ok that makes sense. We will have to modify the approach slightly for `animation` though to accept multiple animations since the animation shorthand supports this:

`animation: scale 1.0s ease-in, slide 2.0s ease`

possibly something like this:

```javascript
const Rotating = styled.div`
  ${animation(['scale', '1.0s', 'ease-in'], ['slide', '2.0s', 'ease'])}
`
```

mxstbr:
That sounds reasonable! I'll see if I can get that implemented 🔥

mxstbr:
Updated, check out the new tests! What do you think about this behaviour?

mxstbr:
Eslint is failing due to this error, but we're inferring that from the types anyway:

```
Missing JSDoc @returns for function  valid-jsdoc
```

I'll disable that rule!

mxstbr:
We can't just disable the `return` stuff, we'd have to disable that rule entirely. What do you think about doing that?

I'm working on the Jest snapshot integration for documentation.js, so the sole thing that'll be left in the JSDoc comments is the description of the utils – everything else should soon be generated out of the code itself. I think it's fine to just disable that rule?

bhough:
> Eslint is failing due to this error, but we're inferring that from the types anyway

It is weird that it is failing, as none of the other modules have an explicit `@returns`. Wonder if that has to do with the branching return statements. My preference is to keep the rule to ensure that even the parts of JSDoc we will still use are present, but if we can't get it to pass any other way...

> What do you think about this behaviour?

Looks good from a behavior standpoint!

Issue #4, opened 2016-11-03
[V2] Error and Warning Handlers


bhough:
Given that this is acting as a CSS compiler of sorts, each module should have error/warning handling that informs the user if they have provided the invalid values/syntax/etc... These should only be use in development and will be stripped out by webpack for production.

In my experiments I've been loosely following the formatting used in [ELM](http://elm-lang.org/blog/compiler-errors-for-humans). So they come out looking sort of like this:

```
-- Warning: Clear Fix -------------------------------------------------------- mixins/ClearFix.js--

  'clearFix' does not accept a parameter. However, you provided a parameter with a value of 50px.

  'clearFix' simply returns the cross-browser CSS necessary to clear floats and does not require
 any parameters to work.

--------------------------------------------------------------------------------------------------
```

Would love everyone's thoughts on this.

mxstbr:
I like having very clear error messages, that could prove important. Let's see what we can do in our build system to strip those out in production but have them in dev!

ericclemmons:
Setting the bar for error messages would be a huge differentiator for this project.

ericclemmons:
@bhough Is there a util that structures output like this already, or something we'd invent?

The last one I looked at was:
> https://github.com/AriaMinaei/pretty-error

I _believe_ we'd be able to throw an object that contained fields like `message` (typical `Error` property) and `hint` that could be leveraged for special output.

mxstbr:
Reusing some common util would be great for this, or writing our own in a separate package!

bhough:
I do like `pretty-error`'s formatting. I'm starting on this issue next, so I will see if it will work with the verbose errors we want to display, if not I like the idea of creating our own @mxstbr.

One of the tricks I wanted to implement with this is to remove this error/warning functionality when the package is built for prod. I imagine that while useful for debugging during development, it would prove to be superfluous code for production. Does that assumption hold up with everyone?

mxstbr:
Yeah that sounds very reasonable to me!

bhough:
@mxstbr @nikgraf Question on this. As far as what errors we want to expose in dev/production. Right now I have formatted errors/warnings working in dev and completely disabled in prod (meaning no custom errors are thrown in prod at all). Are we fine with this, or do we want to draw a specific line between for what we leave in prod when it comes to custom errors.

mxstbr:
That sounds reasonable to me @bhough! Do you have a WIP PR?

nikgraf:
[...]

PR #34, opened 2016-12-08
Add em helper

mxstbr:
Adds the `em` helper. Please check if this works the way you'd expect it to!

I had to add a `eslint-disable` for the parameters reassignment, what do you think about disabling that rule?

Also `stripUnit` might return a string since it returns the original string if it's not a number, so flow complains if I just do `pxval / base` because those _might_ be strings. Unsure what we want to do here, so I've just disabled flow for that line. Any ideas? /cc @relekang


relekang:
>Unsure what we want to do here, so I've just disabled flow for that line. Any ideas?

What about this?
```es
return `${parseInt(pxval, 10) / parseInt(base, 10)}em`
```

mxstbr:
Well but `stripUnit` already does that. The case of a string being there would only ever happen when `base` or `pxval` are invalid.

relekang:
Ah okay, should have checked the source of `stripUnit`. 

I think the flow team is working on improving the applications of typeof checks. We can probably add the suppresser for now an look more into what we can do about it. However, with the suppresser I think we should type the return value of the function.

Using a function and type it with `/* : %checks */` like we did for [`isTag` in styled-components][isTag] might work.

[isTag]: https://github.com/styled-components/styled-components/blob/master/src/utils/isTag.js

mxstbr:
> Using a function and type it with /* : %checks */ like we did for isTag in styled-components might work.

Unsure how that works, but you should be able to push to this branch – would you mind trying that? Thanks!

relekang:
Sure, I do not have time today, but will do it tomorrow ✌️

mxstbr:
Fixed by throwing an error from `em()`, what do you think @bhough? Is this reasonable/expected behaviour?

bhough:
> Fixed by throwing an error from em(), what do you think @bhough? Is this reasonable/expected behaviour?

I think so, once we implement #4 we will be doing a lot this anyway.

> I had to add a eslint-disable for the parameters reassignment, what do you think about disabling that rule?

Maybe better to take a functional approach and reassign it to a new value `newPxval = stripUnit(pxval)`.

src/helpers/em.js
```diff
@@ -0,0 +1,52 @@
...
+  if (typeof pxval === 'string') {
+    pxval = stripUnit(pxval)
+  }
+
+  if (typeof base === 'string') {
+    base = stripUnit(base)
+  }
+
+  if (typeof pxval === 'string') {
+    throw new Error(`Passed invalid pixel value ("${pxval}") to em(), please pass a value like "12px" or 12.`)
```
bhough: Should we handle the case where a valid string value is passed but of a unit that is not px. For example if someone tried to do `em(12rem)`. Since the method name is just `em` it isn't clear immediately (outside of docs) that states it only accepts `px`.

src/helpers/em.js
```diff
@@ -0,0 +1,52 @@
...
+ * element {
+ *   'height': '1em'
+ * }
+ */
+
+/* eslint-disable no-param-reassign */
+function em(pxval: string|number, base?: string|number = '16px') {
+  if (typeof pxval === 'string') {
+    pxval = stripUnit(pxval)
+  }
```
bhough: Maybe a functional approach would be better here where we aren't mutating the original arg.

src/helpers/em.js
```diff
@@ -0,0 +1,52 @@
...
+
+  if (typeof base === 'string') {
+    base = stripUnit(base)
+  }
```
bhough: see above ^

src/helpers/em.js
mxstbr: True, good point! Will add that!

mxstbr:
Resolved all the issues, please review again!

src/helpers/em.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * element {
+ *   'height': '1em'
+ * }
+ */
+
+/* eslint-disable no-param-reassign */
+function em(pxval: string|number, base?: string|number = '16px') {
+  let newPxval = pxval
+  let newBase = base
```
bhough: Should we just initialize these here and then assign them below? No need to assign them to values we don't use in order to just mutate them.

src/internalHelpers/_endsWith.js
```diff
@@ -0,0 +1,6 @@
+/**
```
bhough: Need to add `@flow` here so it will pass CI.

bhough:
@mxstbr Two small comments, otherwise looking good. The `endsWith` helper will be nice for some better error handling on existing mixins as well.

src/helpers/em.js
mxstbr: No, because if both of them are numbers we never assign anything else and those values will stay.

mxstbr:
Added `@flow` to `endsWith`, the assignment actually needs to be there in case numbers and not strings are passed! Ready to go I think?

bhough:
Will need to somehow exclude `internalHelpers` from the JSDoc rules. Otherwise, good to go.

mxstbr:
Just need to mark it as `@private`, one sec!

PR #35, opened 2016-12-08
Add rem helper

mxstbr:
Adds the `rem` helper

This is 100% based on the em helper from #34


mxstbr:
Updated to be exactly like the just merged `em` helper, then realized that they basically look exactly the same. I created a `_pxto` factory which allows us to very easily create those two with almost no code repetition. (except documentation)

Also moved all the tests over there. Feel free to review!

Issue #31, opened 2016-12-08
Doc theme


mxstbr:
We definitely need a nicer theme for the documentation.

The things people will come to the docs site for are:

1. Find a list of supported helpers
2. Find out how to use a certain helper
3. Find out which options a certain helper has

The current theme a) doesn't look very inviting (needs a nicer color scheme, we can probably just take the `styled-components` one) and b) takes actual brain power to understand how to use a certain helper. I like the sidebar, and I like how the parameters are documented though. We should get rid of the `Returns` section though, as it's totally useless for our purposes.

For reference, here are the [Bourbon docs](http://bourbon.io/docs/) which are pretty clear and easy to understand:

![screen shot 2016-12-08 at 16 20 33](https://cloud.githubusercontent.com/assets/7525670/21015465/5b4d1eea-bd62-11e6-8ed0-d57eef301b81.png)

I like their ideas for the header, instead of our massive violet wall of color and text it should also just say "Ellipsis" on the left side and link to the source on the right side.

Maybe @okonet has some ideas for this?

mxstbr:
Something that would be really cool is to have live REPLs (maybe embedded JSBins/JSFiddles/Codepens/...) which show the helper usage on the left, and the live output what it returns on the right. That way people coming to the docs can only see an example of usage but also play around with it! 😱 💯

mxstbr:
We should also use a syntax highlighter that supports ES6. (the `${timingFunction}` here should be highlighted as JS code again)

![screen shot 2016-12-08 at 16 48 18](https://cloud.githubusercontent.com/assets/7525670/21016535/2bd8e7bc-bd66-11e6-8e95-649e47fec465.png)

mxstbr:
Should we switch to using [documentation.js](https://www.npmjs.com/package/documentation)? They support Flow, meaning we don't duplicate any information between the comments and the flow types which is super rad, ES7+ and has a GitHub integration – in general it just sounds much better than JSDoc!

I haven't really dug in, but what do you think @bhough? Do you reckon we should do that or is our JSDoc setup so intricate it will take ages to replicate it?

bhough:
@mxstbr I believe it works with JSDoc syntax, so superficially it should be a fairly easy transitions. How the template would hold up would be another story. However, if someone picks up this up, that would be moot.

mxstbr:
Switched to documentation.js, inlining the default theme with #54! Would be 👌 to get some design for the homepage and the docs page

okonet:
[...]

PR #37, opened 2016-12-08
Integrate documentation.js

mxstbr:
This is so nice! Looks like a huge PR, but is actually much less code now since the theme is no longer inlined. I was able to remove almost all JSDoc comments (except for the examples) in favor of flow types, and documentation understood everything perfectly! 😱 So super cool!

This is what it looks like right now, which to be honest I quite like!

screen shot 2016-12-08 at 21 16 19

Whatcha think @bhough?

Ref: #31


bhough:
@mxstbr this looks great! Looks like it will need #38 to pass CI, with that and conflict resolution let's merge it ASAP.

mxstbr:
Let's 🚢 this first thing @bhough! Can you review & merge?

package.json
```diff
@@ -55,6 +55,7 @@
     "babel-preset-latest": "^6.16.0",
     "cz-conventional-changelog": "^1.2.0",
     "docdash": "^0.4.0",
+    "documentation": "^4.0.0-beta16",
```
bhough: Should we remove the JSDocs and associated Babel plugin as a dependency now?

mxstbr: Arg yes we should, good catch! Mind doing that and pushing to the branch?

PR #38, opened 2016-12-09
Disable max-len rule

mxstbr:
This rule is just unnecessarily annoying and makes typing things really
hard. I'm in favor of just straight up disabling it!

@bhough what do you think?


mxstbr:
All current PRs are solely failing due to this rule, so when this is merged feel free to go through and update all failing PR branches!

Issue #40, opened 2016-12-09
Fix hiDPI export


mxstbr:
Right now `index.js` says this for the export:

```JS
export {
  hiDIPI
}
```

Should be `hiDPI`, not `hiDIPI`!

bhough:
Closed by #42

Issue #33, opened 2016-12-08
Word Wrap misses styled-components example


mxstbr:
The wordWrap mixin is missing a `styled-components` example:

![screen shot 2016-12-08 at 16 47 47](https://cloud.githubusercontent.com/assets/7525670/21016510/17d8a342-bd66-11e6-9dfa-101c9fc73d12.png)

bhough:
Closed by #42

PR #41, opened 2016-12-10
Adds selection mixin

bhough:
Adds selection mixin.

Also addresses issues in #33 and #40.

### DIFF
--- a/src/mixins/selection.js
+++ b/src/mixins/selection.js
@@ -1,4 +1,33 @@
 // @flow
+
+/**
+ * CSS to style the selection psuedo-element.
+ *
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   ...selection({
+ *     'background': 'blue'
+ *   }, 'section')
+ * }
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   ${selection({'background': 'blue'}, 'section')}
+ * `
+ *
+ * // CSS as JS Output
+ *
+ * 'div': {
+ *   'section::-moz-selection': {
+ *     'background-color':'blue',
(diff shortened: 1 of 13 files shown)


src/mixins/selection.js
```diff
@@ -0,0 +1,42 @@
...
+ *   'section::-moz-selection': {
+ *     'background-color':'blue',
+ *   },
+ *   'section::selection': {
+ *     'background-color': 'blue',
+ *   }
+ * }
+ */
+
+function selection(styles: Object, parent: string = '') {
```
bhough: I lean towards no on this one. Unlike other pseudo selectors `::selection` is allowed and often the most common use case, where other pseudos need a parent to function.

src/mixins/selection.js
```diff
@@ -0,0 +1,42 @@
...
+ * // Styles as object usage
+ * const styles = {
+ *   ...selection({
+ *     'background': 'blue'
+ *   }, 'section')
+ * }
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   ${selection({'background': 'blue'}, 'section')}
```
bhough: My flow-fu isn't super strong. Setting `styles` to `Object|Array` throws this error `Incorrect number of type parameters (expected 1)`. Is there a specific syntax beyond `any` when you want a param to be one of two types?

mxstbr: Uhh, I don't actually know. It might need to be `Object|Array<any>`? Have you tried that? /cc @relekang

mxstbr: Just realized that this won't work anyway, so this is good to go!

PR #42, opened 2016-12-10
Add placeholder mixin

bhough:
Add placeholder mixin


src/mixins/placeholder.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   ...placeholder(styles)
+ * }
+ *
+ * // styled-components usage
+ * const div = styled.input`
+ *   ...placeholder(styles)
```
mxstbr: No `...`, this should be `${placeholder(css`background: red`)} `

Since `css` returns an Array, you'll also need to amend the types to allow for that!

bhough: Would `placeholder` need its types amended though? It's` css` that is modifying `placeholder`'s output to an array. If another JS to CSS library implemented another means that changed the type we would need to add that too.

bhough:
> I'm not sure if we should add this – shouldn't the autoprefixers of the various libraries take care of that?

When we originally made the list I had thrown together a bunch of prefixers for testing and came across a few that did not catch this or `::selection`.  I can go back through them and see if that is still the case. Could be useful to leave this and #41 in and deprecate later if we think that changes. Open to discussion on this though.

mxstbr:
Okay, that sounds reasonable! Thanks for checking that!

src/mixins/placeholder.js
mxstbr: No,  the input is then an `array`,  the output is still an object.

bhough: duh...coding past midnight is never a good idea. I'll fix this in the morning.

mxstbr: Any update here?

bhough: just got back on this. I want to make sure I understand this situation, because it is more than just amending the types from what I can see. Here is what I'm seeing:

```javascript
const styles = {
  "background": "blue",
  "color": "red"
};

return css`${styles}`
```

This returns
```javascript
[
"background: blue; color: red;"
]
```

That means we would be passing `placeholder` an array with a single item that contains all the styles passed into it. Is that correct?

If that is the case I will need to refactor this, as spreading the passed in styles would not accommodate this use case.

mxstbr: That should totally work for styled components:

```JS
const Box = styled.div`
  ${placeholder(css`styles`)}
`
```

We support arrays too, so this should work perfectly fine if just passed through, but please try it out!

bhough:
@mxstbr this is ready...again ;)

PR #43, opened 2016-12-10
Add size mixin

bhough:
Add size mixin, docs, and associated tests.


src/mixins/size.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+/**
+ * CSS to represent truncated text with an ellipsis.
```
mxstbr: Fix description!

PR #45, opened 2016-12-13
Add the modularScale helper

mxstbr:
~~Adds the `modularScale` helper using the lovely `simple-modular-scale` package by jxnblk. _(not tagged so he doesn't get annoyed by the notification)_~~

~~Also implements caching the computed scales so we don't recompute the scale everytime somebody requests the same one with different steps. I have not benchmarked this or anything, but essentially the second time somebody gets a value from a scale with the same base and ratio the function is just an array lookup; that should be much cheaper than recomputing the entire scale for every single value! :tada:~~

I was talking absolute bullshit, we don't need a module for this, simply doing `base * ratio * steps` is good enough. Fixed and vastly simplified in the last commit!


mxstbr:
CI is failing due to these totally unnecessary rules since they are actually documented in the flow types:

```
  49:1  error  Missing JSDoc @returns for function  valid-jsdoc
  49:1  error  Missing JSDoc for parameter 'steps'  valid-jsdoc
  49:1  error  Missing JSDoc for parameter 'base'   valid-jsdoc
  49:1  error  Missing JSDoc for parameter 'ratio'  valid-jsdoc
```

Should we just disable that eslint rule and merge it into master to fix this?

mxstbr:
Uhh, something went really wrong here with the commits 😱 any idea how we can resolve this? I obviously did something wrong 😕

bhough:
You could cherry pick your modular scale commits onto a new branch from master and do a new pull request. Would probably be the easiest way. If you wanted to keep this PR you could force reset it to master, and  then cherry pick those commits back onto it.

mxstbr:
Replaced by #48!

PR #48, opened 2016-12-14
Add modularScale

mxstbr:
Replaces #45, adds `modularScale` helper!


mxstbr:
This does not currently show the possible ratios in the documentation:

screen shot 2016-12-16 at 11 20 50

Opened an issue upstream to add support for this (https://github.com/documentationjs/documentation/issues/632), I'm hoping this will be possible. Otherwise we'll have to resort to JSDoc for this util for launch.

PR #53, opened 2016-12-15
Rgb

nikgraf:
The one thing missing is to reduce the hex values e.g. `#ffffff` to `#fff`. I will add this sometime later.


src/helpers/rgb.js
```diff
@@ -0,0 +1,48 @@
...
+
+type RgbColor = {
+  red: number,
+  green: number,
+  blue: number,
+}
+
+function toHex(value) {
+  const hex = value.toString(16)
+  return hex.length === 1 ? `0${hex}` : hex
```
mxstbr: When is `hex` ever just one character and why do we prefix 0 in that case?

mxstbr:
Looks amazing, just the short hex notations and then let's :ship:

src/helpers/rgb.js
nikgraf: If you create the hex of 0 it will be `'0'`. The issue here at hand is that only of 3 parts of a hex color can be reduced only then it should be. Otherwise two characters e.g. prefixed with a `0` are required.

bhough:
Great to see the color portion of this underway, thanks for the contribution @nikgraf!

PR #54, opened 2016-12-15
Inline documentation theme, add homepage

mxstbr:
This PR inlines the documentation theme into the repo so we can edit it. Also moves the actual documentation to `/docs` and a separate thing as the homepage so we can have a real homepage!


mxstbr:
For some reason I managed to get rid of the `docs/assets/` folder – don't merge this yet, need to fix that first!

bhough:
@mxstbr approved, I will let you merge when you're ready.

mxstbr:
Let's ship this now and iterate as we go, including #55!

PR #55, opened 2016-12-15
Autogenerate examples from tested code

mxstbr:
*Note: this builds upon #54, merge that first*

This PR adds the ability to auto-generate examples in the documentation from `examples.js` files! 🎉

Currently, I've only added one to demostrate this. This is what the `stripUnit.examples.js` file looks like:

```JS
import stripUnit from '../stripUnit'

export default [
  { '--dimension': stripUnit('100px') },
  `
  --dimension: ${stripUnit('100px')}
`,
]
```

I've removed all the `@example` stuff from the `stripUnit.js` itself, it now only has the high-level explanation in the comment:

```JS
/**
 * Strip the unit from a given CSS value, returning just the number. (or the original value if an invalid string was passed)
 */

function stripUnit(value: string): number|string {
   /* ... */
}
```

This is what the generated documentation looks like:

screen shot 2016-12-15 at 18 20 02

🎉 🎉 🎉

On top of that, all `.examples.js` files are ran through Jest and snapshotted – meaning the examples in the docs are unit tested and _cannot_ break without us noticing! They literally have to work!

----

Note: This is quite finnicky and will probably break in some cases, but to be honest I'm totally fine with that. We'll continue to improve this as we go along and reach edge cases, but this should easily suffice for our 80% use case.

To see the specific changes (apart from #54) that make this possible look at c3480d966e58f96713c0977746b2bb7acbab0c9c for the test-example-files and 68d92f9f494b98b1cc63d7ae78cd87ac36ef81c2 for the generation-from-example part.

Please comment below if you'd like this to land, then I'll start converting all the utils from comments to `.examples.js` files!


mxstbr:
This is harder than I initially anticipated, there's a bunch of issues with this. (objects getting `toString`ed,...) It's 70% done, but let's leave this as-is for now and not block the release.

PR #56, opened 2016-12-15
Border, Padding, and Margin shorthands

bhough:
Shorthand helpers for border-width, border-color, border-style, margin, padding, and the general directionalProperty helper.


bhough:
@mxstbr Wanted to get your eyes on this implementation before I wrote up the docs for it.

mxstbr:
Looks really good from a first glance! 😱

src/shorthands/position.js
```diff
@@ -0,0 +1,61 @@
...
+ * div {
+ *   'position': 'absolute',
+ *   'top': '12px',
+ *   'right': '24px',
+ *   'bottom': '36px',
+ *   'left': '48px'
+ * }
+ */
+
+function position(positionKeyword: string|null, ...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/shorthands/padding.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * div {
+ *   'padding-top': '12px',
+ *   'padding-right': '24px',
+ *   'padding-bottom': '36px',
+ *   'padding-left': '48px'
+ * }
+ */
+
+function padding(...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/shorthands/margin.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * div {
+ *   'margin-top': '12px',
+ *   'margin-right': '24px',
+ *   'margin-bottom': '36px',
+ *   'margin-left': '48px'
+ * }
+ */
+
+function margin(...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/helpers/directionalProperty.js
```diff
@@ -0,0 +1,51 @@
...
+const positionMap = ['top', 'right', 'bottom', 'left']
+
+function generateProperty(property: string, position: string) {
+  if (!property) return position
+  const splitPropertyName = property.split('-')
+  splitPropertyName.splice(1, 0, position)
+  return splitPropertyName.join('-')
+}
+
+function generateStyles(property: string, valuesWithDefaults: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/shorthands/borderColor.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * div {
+ *   'border-top-color': 'red',
+ *   'border-right-color': 'green',
+ *   'border-bottom-color': 'blue',
+ *   'border-left-color': 'yellow'
+ * }
+ */
+
+function borderColor(...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/shorthands/borderStyle.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * div {
+ *   'border-top-style': 'solid',
+ *   'border-right-style': 'dashed',
+ *   'border-bottom-style': 'dotted',
+ *   'border-left-style': 'double'
+ * }
+ */
+
+function borderStyle(...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/shorthands/borderWidth.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * div {
+ *   'border-top-width': '12px',
+ *   'border-right-width': '24px',
+ *   'border-bottom-width': '36px',
+ *   'border-left-width': '48px'
+ * }
+ */
+
+function borderWidth(...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

src/helpers/directionalProperty.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * div {
+ *   'padding-top': '12px',
+ *   'padding-right': '24px',
+ *   'padding-bottom': '36px',
+ *   'padding-left': '48px'
+ * }
+ */
+
+function directionalProperty(property: string, ...values: Array<string|null>) {
```
mxstbr: If you set `null` as a type flow will think that's an `any`, which we need to avoid. Instead of saying that, do `param: ?string`. That'll tell flow that the param is null-able!

bhough:
@mxstbr should be good now.

mxstbr:
I'm seeing some totally unrelated commits we already have in master here? 

screen shot 2016-12-17 at 21 31 41

bhough:
Odd I'm not seeing those, let me re-push the rebase onto master and see if that fixes it for ya.

bhough:
k @mxstbr did another force push of my branch. I've had some wonkiness going on with my pushes today, as I'm working from Taiwan and the connection isn't always stable. Hopefully that fixed it.

mxstbr:
Awesomeeee!!

PR #59, opened 2016-12-16
Fix ToC in docs

mxstbr:
Updates the ToC with all the new utils we have :tada:


mxstbr:
The docs theme uses ES6, which isn't supported in Node 4. (which we test for in Travis) I say we drop Node 4 support for development, people should be using Node 6 on their dev machines anyway, and thusly from Travis!

bhough:
I am going to revert this @mxstbr the docs are completely broken with this PR.

PR #61, opened 2016-12-16
Revert "Fix ToC in docs"

bhough:
Reverts styled-components/polished#59


mxstbr:
No leave it, I'll fix it. One sec!

mxstbr:
They look alright to me? Moved them to `/docs`... https://polished.js.org/docs/

bhough:
Locally they are completely broken now and I see a docs folder with another docs folder inside of it in master, both with index.html files. Is this expected?

mxstbr:
Yes, since `polished.js.org` (the homepage) will have a, well, homepage. That's why I moved the docs to `/docs` 😊 

Go to polished.js.org, nothing there (since there is not homepage yet, going to work on that), go to https://polished.js.org/docs, there's docs! 🎉

bhough:
Ok, would be nice if we can link up the assets so the docs work locally. Right now they don't get styling or the js because the relative paths don't resolve. I

mxstbr:
I use [`pushstate-server`](http://npm.im/pushstate-server) locally like this: `$ pushstate-server docs`, and it works for me. What do you use that it doesn't work?

bhough:
I was just loading up index.html in the browser direction, i can move to using that.

mxstbr:
Ooh, yeah that'll break since I had to resort to absolute links to the assets to make them work for the `docs/` suffix 😕 Maybe we should add an npm script for that?

bhough:
I'll look into adding it as part of #56

PR #63, opened 2017-01-03
chore(font-face): @font-face mixin

bhough:
@font-face mixin, accepts both file-based and local references. Also is configurable for file formats, font options, and unicode-range.

Addresses an item in #1

### DIFF
--- a/src/mixins/fontFace.js
+++ b/src/mixins/fontFace.js
@@ -1,5 +1,34 @@
 // @flow
 
+type ConfigurationType = {
+  fontFamily: string;
+  fontFilePath?: string;
+  fontStretch?: string;
+  fontStyle?: string;
+  fontVariant?: string;
+  fontWeight?: string;
+  fileFormats?: Array<string>;
+  localFonts?: Array<string>;
+  unicodeRange?: string
+}
+
+function generateFileReferences(fontFilePath: string, fileFormats: Array<string>) {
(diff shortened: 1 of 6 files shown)


bhough:
@mxstbr this is ready for review. Travis is timing out on tests only on the push build, but I think it is a hiccup on their end. Especially want your feedback on the examples, since @font-face works differently than traditional CSS properties.

src/mixins/fontFace.js
```diff
@@ -0,0 +1,93 @@
...
+      'unicode-range': unicodeRange,
+      'font-stretch': fontStretch,
+      'font-style': fontStyle,
+      'font-variant': fontVariant,
+      'font-weight': fontWeight,
+    },
+  }
+
+  // Removes undefined fields for cleaner css object.
+  return JSON.parse(JSON.stringify(fontFaceDeclaration))
```
mxstbr: Is there a faster way of doing this?

bhough:
@mxstbr I looked at pruning by looping over with Object.keys and it proved to be about the same speed, with more code. The question is if we are okay with a slightly more opaque approach with a comment vs a more verbose but slightly clear approach (with Object.keys).

src/mixins/fontFace.js
bhough: @mxstbr I looked at pruning by looping over with Object.keys and it proved to be about the same speed, with more code. The question is if we are okay with a slightly more opaque approach with a comment vs a more verbose but slightly clear approach (with Object.keys).

PR #64, opened 2017-01-03
chore(shorthands): Background & Background Image

bhough:
Shorthands for multiple backgrounds and background-images

Addresses 2 items in #1


bhough:
@mxstbr all the sudden seeing a lot of timeouts on travis for the test suite. Even re-running old builds is now taking forever.

PR #68, opened 2017-01-13
chore(borderRadius): borderRadius shorthand

bhough:
Provides shorthand for applying a radius value to both corners of a side.

### DIFF
--- a/src/shorthands/borderRadius.js
+++ b/src/shorthands/borderRadius.js
@@ -1,0 +1,43 @@
+// @flow
+
+/**
+ * The border-radius shorthand accepts a value for side and a value for radius and applies the radius value to both corners of the side.
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   ...borderRadius('top', '5px')
+ * }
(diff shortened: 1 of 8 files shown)


bhough:
@mxstbr rebased and ready review.

src/shorthands/borderRadius.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (side === 'left' || side === 'right') {
+    return {
+      [`border-top-${side}-radius`]: radius,
+      [`border-bottom-${side}-radius`]: radius,
+    }
+  }
+
+  throw new Error('borderRadius expects a valid value for side.')
```
mxstbr: Could this error be a bit more descriptive maybe? "borderRadius expects one of "top", "bottom", "left" or "right" as the first argument."

bhough: Yep I will make this change. Also going to add yarn-error.log to the .gitignore while I'm at it since that is showing up in PRs now.

PR #70, opened 2017-01-14
Hsl

nikgraf:


src/helpers/hsl.js
```diff
@@ -0,0 +1,44 @@
...
+import hslToHex from '../internalHelpers/_hslToHex'
+
+type HslColor = {
+  hue: number,
+  saturation: number,
+  lightness: number,
+}
+
+/**
+ * Returns a string value for the color. The returned result is the smalles possible hex notation.
```
mxstbr: typo: smallesT

src/internalHelpers/_hslToHex.js
```diff
@@ -0,0 +1,53 @@
...
+import toHex from './_numberToHex'
+
+function hslToHex(hue: number, saturation: number, lightness: number): string {
+  if (saturation === 0) {
+    // achromatic
+    const colorComponent = lightness * 255
+    return reduceHexValue(`#${toHex(colorComponent)}${toHex(colorComponent)}${toHex(colorComponent)}`)
+  }
+
+  // formular from https://en.wikipedia.org/wiki/HSL_and_HSV
```
mxstbr: typo: formula (no "r")

src/internalHelpers/_hslToHex.js
```diff
@@ -0,0 +1,53 @@
...
+    green = 0
+    blue = chroma
+  } else if (huePrime >= 5 && huePrime < 6) {
+    red = chroma
+    green = 0
+    blue = secondComponent
+  }
+
+  const lightnessModification = lightness - (chroma / 2)
+  const addLightAndConvert = (color) => toHex(Math.round((color + lightnessModification) * 255))
```
mxstbr: Should we extract this as a separate util, will it be useful in any other helpers? 

At least we should extract it out of this function to avoid allocating an unnecessary variable every time `hsl` is called?

nikgraf: Good point, I just extracted it. I think it will only be used in hslToHex as it's specific to the hslToRgb formula, but once I see us using it twice I will extract it 👍

PR #74, opened 2017-01-15
Hsla

nikgraf:


src/helpers/hsla.js
```diff
@@ -0,0 +1,63 @@
...
+
+type HslColor = {
+  hue: number,
+  saturation: number,
+  lightness: number,
+  alpha: number,
+}
+
+/**
+ * Returns a string value for the color. The returned result is the smallest possible hex notation.
```
mxstbr: This doesn't make any sense, "is the smallest possible hex notation"? It's actually converted to `rgba` with a hex fallback, right? Hex notation can't have transparency...

nikgraf: Good catch, just fixed it

PR #75, opened 2017-01-16
Add missing flow type references

### LINKED ISSUE #71: Flow type references missing in docs
Hi, guys,

I see some flow type references missing in docs like this `RgbaColor`.

image

I'm playing with `Flow` and `Documentation` on another project and also faced this problem, but I finally managed to solve it.

Would a PR be welcome?

diegohaz:
For some reason, adding a doc block to the type solves this. I did change some names for consistency and to avoid location hash conflicts (the `TimingFunctions` type and `timingFunctions` method had the same id, so I changed type to `TimingFunction`).

Also, `documentation` still does not recognize some complex types like `Ratio` but it links to the actual code, so it's ok.

Closes #71

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -52,6 +52,19 @@
     "em",
     "modularScale",
     "rem",
-    "stripUnit"
+    "stripUnit",
+    {
+      "name": "Types",
+      "description": "Types used within the library."
+    },
+    "Ratio",
+    "RgbColor",
+    "RgbaColor",
+    "FontFaceConfiguration",
+    "RadialGradientConfiguration",
+    "TimingFunction",
+    "AnimationProperty",
+    "ButtonState",
+    "InputState"
   ]
 }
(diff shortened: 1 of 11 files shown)


mxstbr:
Hmm, since these types correlate 1:1 with the utils that use them, wouldn't it make more sense to show them next to e.g. the function signature there? Instead of as their own full-blown item?

diegohaz:
If we don't add them to documentation.json, they will be added automatically at the end of toc. We can move them next to their functions, but they will still appear on toc. Between those two options, I think it's better to group types than mixing types and functions.

But there may be a config or some jsdoc tag to handle this I'm not aware of.

mxstbr:
Hm, I'd much prefer them to not be their own items but live with the util they belong to. (not even mentioned in the ToC, that is)

I was just going to see if documentationjs has an open issue about this somewhere, and it turns out I submitted one 😁 https://github.com/documentationjs/documentation/issues/632

I guess in that case your solution is better than nothing, thanks!

PR #77, opened 2017-01-16
chore(triangle): Add triangle mixin

machycek:
Added triangle mixin with tests.

My first PR, hope it helps.


src/mixins/triangle.js
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @example
+ * // Styles as object usage
+ *
+ * const styles = {
+ *   ...triangle({ pointing: 'right', width: '100px', height: '100px'})
+ * }
+ * const styles = {
+ *   ...triangle({ pointing: 'right', width: '100px', height: '100px'})
+ * }
```
mxstbr: Duplicate example

src/mixins/triangle.js
```diff
@@ -0,0 +1,117 @@
...
+  `${height}px 0 ${height}px ${width}px`
+)
+
+const getTopPointingTriangle = ({ height, width } : TriangleSizes) => (
+  `0 ${width}px ${height}px ${width}px`
+)
+
+const getBottomPointingTriangle = ({ height, width } : TriangleSizes) => (
+  `${height}px ${width}px 0 ${width}px`
+)
```
mxstbr: What about refactoring this to a single function? Something like

```JS
const getBorder = ({ height, width, direction }): TriangleSizes) => {
  switch (direction) {
    case 'left': return `${height}px ${width}px ${height}px 0`
    case 'right': return `${height}px 0 ${height}px ${width}px`
    case 'top': return `0 ${width}px ${height}px ${width}px`
    case 'bottom': return `${height}px ${width}px 0 ${width}px`
  }
}
```

This would allow us to get rid of a bunch of `if`s in the main function. 😊

src/mixins/triangle.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+
+  // needed for border-color
+  const reverseDirection = {
+    left: 'right',
+    right: 'left',
+    top: 'bottom',
+    bottom: 'top',
+  }
```
mxstbr: Since this doesn't ever change, let's move it outside of the function so we don't allocate unnecessary memory everytime `triangle` is used.

machycek:
Should be fixed now, cheers for feedback 👏🏼

src/mixins/triangle.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+/* eslint-disable default-case, consistent-return */
```
mxstbr: Instead of disabling these, let's throw an error if a direction other than `top/bottom/left/right` is passed.

In the `switch` add a `default` statement that throws a helpful error so the user can fix their code!

machycek:
Hopefully ready now 👍🏼

mxstbr:
Now we've got conflicts! 😕 This should solve it:

```sh
git merge master # Don't forget to pull the newest changes into master first!
# There'll be conflicts, which are easy to resolve just by doing:
npm run build:docs
# Now finish the merge
git commit
# and push to this repo!
```

machycek:
How about now?

machycek:
Awesome, cheers 👍🏼 started working on Opacity - transparentize 😊

PR #80, opened 2017-01-31
Add mix, hex2rgb, hex2rgba and alpha.

ahmed3mar:
Hope it help :)


bhough:
Thanks for the PR @ahmed3mar. I will take a look at it this weekend.

src/internalHelpers/_rgbToHSL.js
```diff
@@ -0,0 +1,31 @@
+const rgbToHSL = (rgba) => {
```
nikgraf: This is suuuuper useful!

src/helpers/mix.js
```diff
@@ -0,0 +1,66 @@
...
+  if (typeof weight === 'string') {
+    if (endsWith(weight, '%')) {
+      newWeight = stripUnit(weight)// newWeight.substr(-1)
+    }
+  }
+  if (newWeight > 100) newWeight = 100
+
+  const c1 = new Color(newColor1)
+  const c2 = new Color(newColor2)
+  const newColor = new Color('mix', [c1, c2, newWeight])
```
nikgraf: The idea of the existing design was to only leverage functions, because then we can slit them up quite nice and people only need to import the most necessary part. By using a Color class and moving utils there we only can do it when relying on tree-shaking that removes Class methods.

.documentation.json
```diff
@@ -28,6 +28,9 @@
     "rgba",
     "hsl",
     "hsla",
+    "mix",
+    "hex2rgb",
+    "hex2rgba",
```
nikgraf: I think we shouldn't expose hex2rgb & hex2rgba. What's the benefit for the user here? When would she/he use it?

ahmed3mar: idk but i love to add all possible options to user

ahmed3mar:
@bhough Thanks ^^

@nikgraf Yup i'll try to spit my PR to use it for minimal API and add more testing too =D

nikgraf:
@ahmed3mar I'm closing this PR. I would be still happy if you contribute 👍

PR #84, opened 2017-02-12
Color: Darken & Lighten

nikgraf:


mxstbr:
Still WIP or ready for review?

nikgraf:
@mxstbr now ready for review 👍

README.md
```diff
@@ -73,13 +75,16 @@ A lightweight toolset for writing styles in Javascript. ✨
   <ul>
     <li><a href="http://polished.js.org/docs/#ratio">Ratio</a></li>
     <li><a href="http://polished.js.org/docs/#rgbcolor">RgbColor</a></li>
+    <li><a href="http://polished.js.org/docs/#rgbcolor-1">RgbColor</a></li>
+    <li><a href="http://polished.js.org/docs/#rgbcolor-2">RgbColor</a></li>
```
mxstbr: Why do we have three different `RgbColor` types?

src/helpers/darken.js
```diff
@@ -0,0 +1,43 @@
...
+ *   background: "#ffbd31";
+ *   background: "rgba(255,189,49,0.7)";
+ * }
+ */
+function darken(amount: number, color: string): string {
+  // parse color string to hsl
+  const hslColor = rgbToHsl(parseColorString(color))
+  const value = {
+    ...hslColor,
+    lightness: Math.max(0, Math.min(1, hslColor.lightness - amount)),
```
mxstbr: `Math.max(Math.min))` stuff could be extracted into a common `guard(0, 1, val)` helper, do we need it anywhere else?

src/internalHelpers/_parseColorString.js
```diff
@@ -0,0 +1,45 @@
...
+  green: number,
+  blue: number,
+  alpha?: number,
+}
+
+const hexRegex = new RegExp('^#[a-fA-F0-9]{6}$')
+const reducedHexRegex = new RegExp('^#[a-fA-F0-9]{3}$')
+// eslint-disable-next-line no-useless-escape
+const rgbaRegex = /^rgba\((\d+),(\d+),(\d+),([-+]?[0-9]*[.]?[0-9]+)\)$/
+// const rgbaRegex = new RegExp('^rgba\((\d+),(\d+),(\d+),([-+]?[0-9]*[.]?[0-9]+)\)$')
```
mxstbr: Why is this here?

src/internalHelpers/_parseColorString.js
```diff
@@ -0,0 +1,45 @@
...
+type RgbColor = {
+  red: number,
+  green: number,
+  blue: number,
+  alpha?: number,
+}
+
+const hexRegex = new RegExp('^#[a-fA-F0-9]{6}$')
+const reducedHexRegex = new RegExp('^#[a-fA-F0-9]{3}$')
+// eslint-disable-next-line no-useless-escape
```
mxstbr: Why do we need to disable that rule? There's something in the regex below that shouldn't be escaped, let's not escape it?

src/internalHelpers/_parseColorString.js
nikgraf: Ahh, good catch!
First I used `new RegExp` syntax, but there linting complained, but the escapes were not useless. They were needed. With `/<regex>/` syntax it's detected properly … let's use this and remove the eslint disable

src/helpers/darken.js
nikgraf: sounds good, added a guard helper

nikgraf:
@diegohaz with this PR we moved the color types to a separate type folder. Any ideas how to make the documentation recognise them?

diegohaz:
@nikgraf The empty doc blocks worked for me. I've also put that on a separate file (https://github.com/diegohaz/styled-theme/blob/master/src/types.js)

This is really strange, but I would try to remove the empty line between `//@flow` and the first doc block. 😆

nikgraf:
@diegohaz thanks for the quick reply. will try that …

PR #85, opened 2017-02-14
Modifiers

nikgraf:


src/helpers/desaturate.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+import parseColorString from '../internalHelpers/_parseColorString'
+import rgbToHsl from '../internalHelpers/_rgbToHsl'
+import hslToHex from '../internalHelpers/_hslToHex'
+import hslToRgb from '../internalHelpers/_hslToRgb'
+import guard from '../internalHelpers/_guard'
+
+/**
+ * Returns a string value for the desaturated color.
```
mxstbr: This is a nitpick, but I feel like this description should explain what desaturation is rather than say that it returns a string. That's obvious from the usage examples and the type signature anyway. (same goes for saturation)

src/helpers/desaturate.js
```diff
@@ -7,7 +7,9 @@ import hslToRgb from '../internalHelpers/_hslToRgb'
 import guard from '../internalHelpers/_guard'
 
 /**
- * Returns a string value for the desaturated color.
+* Decreases the intensity of a color. Its range is between 0 to 1. The first
+* argument of the desaturate function is the amount by how much the color
+* intensity should be decreased.
```
mxstbr: Nit: add missing whitespace again

PR #87, opened 2017-02-17
Color functions

nikgraf:
Should only be merged after PR https://github.com/styled-components/polished/pull/85 is merged.

### DIFF
--- a/src/helpers/invert.js
+++ b/src/helpers/invert.js
@@ -5,7 +5,7 @@ import rgb from './rgb'
 import rgba from './rgba'
 
 /**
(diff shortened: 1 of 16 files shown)


src/helpers/invert.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import parseColorString from '../internalHelpers/_parseColorString'
+import rgb from './rgb'
+import rgba from './rgba'
+
+/**
+ * Returns the inverted color meaning the red, green and blue values are invert.
```
mxstbr: That's a pretty redundant description 😉 What about `"Inverts the red, green and blue values of a color"` or something like that?

PR #88, opened 2017-02-17
Invert

nikgraf:
WIP, do not merge yet

### DIFF
--- a/src/helpers/invert.js
+++ b/src/helpers/invert.js
@@ -1,0 +1,18 @@
+// @flow
+
+import parseColorString from '../internalHelpers/_parseColorString'
+import rgb from './rgb'
+import rgba from './rgba'
+
+/**
+ * Returns the inverted color.
+ */
+function invert(color: string): string {
+  // parse color string to hsl
+  const value = parseColorString(color)
+  return value.alpha === undefined || value.alpha >= 1
+    ? rgb(255 - value.red, 255 - value.green, 255 - value.blue)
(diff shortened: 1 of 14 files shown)


src/helpers/invert.js
```diff
@@ -0,0 +1,18 @@
...
+ * Returns the inverted color.
+ */
+function invert(color: string): string {
+  // parse color string to hsl
+  const value = parseColorString(color)
+  return value.alpha === undefined || value.alpha >= 1
+    ? rgb(255 - value.red, 255 - value.green, 255 - value.blue)
+    : rgba(255 - value.red, 255 - value.green, 255 - value.blue, value.alpha)
+}
+
```
nikgraf: @ryyppy can you help me out with Flow here?

It complains about alpha not being available because RgbColor doesn't have an alpha. I thought by checking on undefined it would solve it, but it still complains …
```
> flow check

src/helpers/invert.js:13
 13:   return value.alpha === undefined || value.alpha >= 1
                                           ^^^^^^^^^^^ property `alpha` of unknown type. This type cannot be compared to
 13:   return value.alpha === undefined || value.alpha >= 1
                                                          ^ number

src/helpers/invert.js:15
 15:     : rgba(255 - value.red, 255 - value.green, 255 - value.blue, value.alpha)
                                                                      ^^^^^^^^^^^ property `alpha` of unknown type. This type is incompatible with the expected param type of
 32: function rgba(value: RgbaColor | number, green?: number, blue?: number, alpha?: number): string {
                                                                                     ^^^^^^ number. See: src/helpers/rgba.js:32
```

PR #89, opened 2017-02-18
applyTransform function

10xjs:
Add `applyTransform` helper function to allow shorthand definition of transform property functions. 

Properties that match transform function names are removed from the style and moved into the `transform` property.

It works as a helper to transform an entire block of styles or as a mixin.

Example use:

```js
applyTransform({
  color: red,
  translateX: 10,
  rotate: '15deg',
})
```

Yields:

```css
element {
  color: red;
  transform: translateX(10px) rotate(15deg);
}
```

### TODO:
 - [ ] Rebuild docs

### DIFF
--- a/src/helpers/applyTransform.js
+++ b/src/helpers/applyTransform.js
@@ -1,0 +1,82 @@
+// @flow
+
+import contains from '../internalHelpers/_contains'
+import omit from '../internalHelpers/_omit'
+import pick from '../internalHelpers/_pick'
+
+// NOTE: Due to conflict with an actual CSS property, the `perspective`
+// transform function is not included.
+const transformFunctions = [
+  'matrix',
+  'matrix3d',
+  'rotate',
+  'rotate3d',
+  'rotateX',
+  'rotateY',
+  'rotateZ',
+  'scale',
+  'scale3d',
+  'scaleX',
+  'scaleY',
+  'scaleZ',
+  'skew',
+  'skewX',
+  'skewY',
+  'translate',
+  'translate3d',
+  'translateX',
+  'translateY',
+  'translateZ',
+]
+
+// Transform functions which have implicit px units.
+const pxUnits = [
+  'translateX',
+  'translateY',
+  'translateZ',
+]
+
+ /**
+ * Merged transform function properties into the `transform` property. Any
+ * properties on the style object that match css transform functions are removed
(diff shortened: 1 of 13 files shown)


nikgraf:
Nice, the PR is well done!

I'm bit uncertain about the idea. Mixing all the transforms in the root object looks strange to me. I probably would have build a `transform` function. Feels more consistent with the rest of the API, but I'm not sure. @bhough @mxstbr any thoughts?

```js
const styles = {
  transform: transform({
    translateX: 10,
    rotate: '15deg',
  }),
  color: 'red',
}
```

bhough:
Thanks for the PR @10xjs.

I agree with @nikgraf I'm not big on the idea of mixing other root level properties into the mixin. It can get especially messy if they want to use another mixin in parallel. 

I'm not opposed to a standalone transform shorthand though. I know @mxstbr talked about having shorthand for all these compound properties allowing them to support objects. So I imagine this would fall under that.

10xjs:
@nikgraf @bhough Fair criticism. This functionality could definitely be reduced to omit the virtual-property support.

I assume you have been talking about an interface like this:
```
const style = {
   transform: { translateY: 10, rotate: '15deg' },
};
```

If that is something you want to see this turn into, I can refactor things.

Otherwise just close this one :)

nikgraf:
@10xjs that's a good point, hmmm we could so something similar to what we did with color and expose each manipulation as a separate function. The API could look like this:

```js
const style = {
  transform: translateX("20px", "translateX(10px) rotate(15deg)")
}
```

Resulting in:

```js
const style = {
  transform: "translateX(30px) rotate(15deg)"
}
```

10xjs:
@nikgraf 🤔  That doesn't really provide much value beyond the baseline.

```js
import {translateX} from 'polished';

const style = {
  transform: translateX(`${someValue}%`),
}
```

```js
const style = {
  transform: `translateX(${someValue}%)`,
}
```

nikgraf:
agreed!

nikgraf:
@10xjs will close this, let me know if you are interested to help out more

10xjs:
@nikgraf I'm definitely interested in helping out any way I can

nikgraf:
There is a list of chores I want to do: any of those interesting to you? https://github.com/styled-components/polished/issues/91

Issue #44, opened 2016-12-12
Color API


nikgraf:
#I told @mxstbr to help out here and especially with the color related API 🙌 Before starting to code I want figure out how you think about the API. There was already some discussion here https://github.com/styled-components/polished/pull/24, but there are a couple things unclear to me. This is a proposal. Please correct me if I'm wrong or you see room for improvement.

- We want everything color related in the lib to make sure it's properly flow-typed & tested.
- The output of every color related function should be a string.
- We want to create the smallest possible output. For colors without opacity we use hex e.g. `#FF553A`. If possible only 3 characters `#FFF` instead of `#FFFFFF`. In case of opacity we generate `rgba(255,0,0,0.3)`.
- A function might take a string or multiple values e.g. `rgb('22, 255, 0')` and `rgb(22, 255, 0)` and `rgb({ red: 22, green: 255, blue: 0 })`. Not sure about the object structure. Looks handy, but usually I prefer to have one recommended way of doing things and by exposing less possible ways we would enforce it. (We can always add it later)
- Functions like `saturate` accept hex strings or rgb or rgba or hsla strings e.g. `'#FFDDFF'` or `'rgb(22, 255, 0)'`, but don't accept plain single values like rgb (there is no conversion). Otherwise it could get confusion if they accept hsla or rgba. On the other hand if we could accept the object notation, but as said it's not necessary.
- In case of manipulation functions we use a functional style like Rambda with the first argument/arguments being the parameters and the last one the base color. Due the nature of the this API design composition is possible e.g.

```js
// nesting of function calls
const prettyRed = saturate(20, darken(10, rgb(100, 0, 0)));
const prettyBlue = saturate(20, darken(10, rgb(0, 100, 0)));

const prettify = compose(
  darken(10),
  saturate(20)
);
const prettyRed = prettify(rgb(100, 0, 0));
const prettyBlue = prettify(rgb(0, 100, 0));
```

Note: please ignore 10 & 20 arguments. I need to learn more about color theory first to understand what parameters make sense.

What do you guys think? @bhough @mxstbr @guzart

mxstbr:
I love this!!

Two discussion points:

I think we should accept the object structure too, some people prefer that and this shouldn't be the most opinionated lib – it's not like that hurts us. (also, `jss` has added support for object syntax for almost everything so people might be passing those in as variables)

[...]

PR #90, opened 2017-02-18
Mixin: Transparentize

machycek:
Features that are available with `transparentize` function 🎉

- [x] **non transparent** color (rgb, hex, hsl) to transparent color
- [x] **transparent** color (rgba, hsla) to "more" transparent color


src/mixins/transparentize.js
```diff
@@ -0,0 +1,57 @@
...
+const getLastColorValue = (color: string) : number => (
+  Number(
+    color.slice(
+      color.lastIndexOf(',') + 1,
+      color.indexOf(')'),
+    ),
+  )
+)
+
+function transparentize(color: string, percentage : number) {
```
nikgraf: Can you make sure it accepts a range from 0 to 1 instead of 0 to 100.

Why?
I believe it's easier to comprehend for people that if the current value is 0.8 that by passing 0.2 it would end up with 0.6. Having an 0.8 alpha and passing in 20 to reduce it to 0.6 feels strange to me. Thats also the reason why saturation, lightness is using 0 to 1 and we should stay consistent across the lib.

Can you also remove the space after percentage.

src/mixins/transparentize.js
```diff
@@ -0,0 +1,57 @@
...
+    const currentTransparency = getLastColorValue(color)
+    const rgb = getValuesTillLastComma(getColorValuesInBrackets(color))
+    const newTransparency = getNewTransparency(currentTransparency, percentage)
+
+    return `rgba(${rgb}, ${newTransparency})`
+  } else if (isHsl(color)) {
+    const hslValues = getColorValuesInBrackets(color)
+    const newTransparency = getNewTransparency(1, percentage)
+
+    return `hsla(${hslValues}, ${newTransparency})`
```
nikgraf: Right now most of the other functions accept only rgba or hex strings as this is the output from our rgb, rgba, hsl & hsla functions. So even hsla outputs rgba as described here https://github.com/styled-components/polished/issues/44

> - We want to create the smallest possible output. For colors without opacity we use hex e.g. `#FF553A`. If possible only 3 characters `#FFF` instead of `#FFFFFF`. In case of opacity we generate `rgba(255,0,0,0.3)`.

Can you make sure transparentize also exports those? The internal helper contain a couple of helpers that does this conversion.

It might make sense to support `hsl`, `hsla`, `hwb` and `lab` ([all valid CSS](https://drafts.csswg.org/css-color/#the-hwb-notation)) in the future, but I'm not sure. There is a tradeoff. Being open to more inputs means increase code size.

src/mixins/transparentize.js
```diff
@@ -0,0 +1,57 @@
...
+
+const getLastColorValue = (color: string) : number => (
+  Number(
+    color.slice(
+      color.lastIndexOf(',') + 1,
+      color.indexOf(')'),
+    ),
+  )
+)
+
```
nikgraf: Can you add a comment like these? https://github.com/styled-components/polished/blob/master/src/helpers/hsla.js#L7-L32

We use it to generate the docs.

src/mixins/transparentize.js
machycek: Got it. I guess you don't want to use percentage then? At the moment everything is based on percentage eg. transparentize('rgba(255, 255, 255, .5)', 20) => rgba(255, 255, 255, .4)

src/mixins/transparentize.js
machycek: Sounds good, going to change that 👍🏼

machycek:
Glad to hear that, thanks Nik!

src/mixins/transparentize.js
machycek: Should I change it from **"percentage"** to **"amount"** then?

nikgraf: @machycek yes please 👍

machycek:
Ok, so I have done following

- moved transparentize from mixins to helpers
- added comments for the **docs**
- changed the range from 0 - 100 to **0 - 1**
- renamed percentage to amount
- changed logic from percentage to **amount**
- changed order of arguments (amount is **first**, color is last) 
- changed output of hsl / hsla to **rgba**

Looking forward for feedback 👍🏼

PR #92, opened 2017-02-20
Machycek transparentize

nikgraf:


src/helpers/transparentize.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+import rgba from './rgba'
+import parseColorString from '../internalHelpers/_parseColorString'
+import guard from '../internalHelpers/_guard'
+
+/**
+ * Decreases the alpha of a color. Its range for the amount is between 0 to 1.
```
mxstbr: Le't change "alpha" to "opacity" here, since not many people will know exactly what the alpha is.

machycek:
Glad it helped guys 🙌🏼 thanks for feedback!

PR #93, opened 2017-02-21
Opacify

nikgraf:


nikgraf:
@mxstbr jep, we could. I was relying on the existing [Sass functions](http://sass-lang.com/documentation/Sass/Script/Functions.html#opacify-instance_method) here since even the future CSS spec has no modifier to manipulate opacity: https://drafts.csswg.org/css-color/#alpha

PR #100, opened 2017-03-03
Tint and Shade Helpers

bhough:
Adds tint and shade helpers. Also fixes some issues with documentation and alphabetizes color imports.

Addresses items from #1.


nikgraf:
Is tint & shade the same as darken/lighten or different? I find it bit confusing:)

bhough:
Best explanation I've read: 

`Tint and shade are effectively mixing with white and black respectively, whereas lighten/darken are manuipulating the luminance channel independently of hue and saturation. The former can produce hue shifts, whereas the latter does not.`

src/color/shade.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import mix from './mix'
+
+/**
+ * Shades a color by mixing it with black. Leverages the mix color module.
```
nikgraf: @bhough can you remove `Leverages the mix color module`? I think it doesn't add much value for a user, but I really liked your explanation in the comment. What if we do something like this:

```
Shades a color by mixing it with black. Compared to `darken` it can produce hue shifts, wheres `darken` manipulates the luminance channel and therefor doesn't produce hue shifts.
```

src/color/shade.js
```diff
@@ -0,0 +1,33 @@
...
+ * // CSS in JS Output
+ *
+ * element {
+ *   background: "#00003f";
+ * }
+ */
+
+function shade(percentage: number, color: string) {
+  if (!percentage || !color) throw new Error('Passed insufficient arguments to shade, please pass a percentage and a color to be shaded.')
+  if (typeof percentage !== 'number' || percentage > 1) throw new Error('Passed an incorrect argument to shade, please pass a percentage less than or equal to 1.')
```
nikgraf: Should we check for a range between -1 to 1? -1 would basically tint the color if I'm not mistaken, but still a valid input.

src/color/shade.js
nikgraf: @bhough can you remove `Leverages the mix color module`? I think it doesn't add much value for a user, but I really liked your explanation in the comment. What if we do something like this:

```
Shades a color by mixing it with black. Compared to `darken` it can produce hue shifts, wheres `darken` manipulates the luminance channel and therefor doesn't produce hue shifts.
```

src/color/shade.js
```diff
@@ -0,0 +1,33 @@
...
+  if (typeof color !== 'string') throw new Error('Passed an incorrect argument to shade, please pass a string representation of a color.')
```
nikgraf: `mix` is checking if it's a color valid to parse. We can drop this check 😄

src/color/tint.js
```diff
@@ -0,0 +1,33 @@
...
+ * // CSS in JS Output
+ *
+ * element {
+ *   background: "#bfbfff";
+ * }
+ */
+
+function tint(percentage: number, color: string) {
+  if (!percentage || !color) throw new Error('Passed insufficient arguments to tint, please pass a percentage and a color to be tinted.')
+  if (typeof percentage !== 'number' || percentage > 1) throw new Error('Passed an incorrect argument to tint, please pass a percentage less than or equal to 1.')
```
nikgraf: See `shade` comments

src/color/shade.js
bhough: I like that, I'll update it.

src/color/shade.js
bhough: I need to update this to capture the negative use case, but I'm not sure I like the idea of allowing either helper to do  the job of the other by accident (putting in a negative number). It feels like it would be better to limit it based on each helpers use case. Thoughts?

nikgraf:
@bhough let's merge this PR. I will be on a plane for 5h and clean it up there. Just want to avoid merge conflicts later on 😄

PR #102, opened 2017-03-05
Color

nikgraf:
@bhough this includes the changes we discussed in the previous PR and some refactoring

Regarding -1 for tint & shade I would suggest we allow it since it is valid to do that. What I think would be cool are warnings that only show up in dev mode and are removed if polished is minified. Basically the same what React does. Then we can make polished accept weird inputs e.g. alpha of 1.1, but warn in dev mode about it. What do you think?


nikgraf:
@mxstbr @bhough I needed to ignore some Flowtype checking in toColorString. I'm confident that this works fine and is rather a flow issue. I probably try to get help from someone of the flow team next week and see if we can fix it.

Should not block a release …

PR #105, opened 2017-03-10
feat(align): added align shorthands

tolgaergin:
Alignment shorthand helps when position element absolutely. For instance;

1)
```javascript
// it puts object to top right corner with -10px margin offset
const Header = styled.div `
  ${align('right', '-10px')}
`
```

it will generate this css
```css
.header {
  position: absolute;
  top: 50%;
  right: -10px;
  transform: translateY(-50%);
}
```

2) 
```javascript
// it puts object to top center
const Header = styled.div `
  ${align('center')}
`
```

it will generate this css
```css
.header {
  position: absolute;
  top: 50%;
  left: 50%;
  transform: translate(-50%, -50%);
  margin-top: 0;
}
```

3)
```javascript
// it puts object to top bottomLeft with 10px margin offset 
const Header = styled.div `
  ${align('bottomLeft', '10px')}
`
```
it will generate this css

```css
.header {
  position: absolute;
  left: 10px;
  bottom: 10px;
}
```

Alignment in styled-components
Alignment in styled-components
Alignment in styled-components
Alignment in styled-components

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,7 @@ import toColorString from './color/toColorString'
 import transparentize from './color/transparentize'
 
 // Shorthands
+import align from './shorthands/align'
 import animation from './shorthands/animation'
 import backgroundImages from './shorthands/backgroundImages'
 import backgrounds from './shorthands/backgrounds'
@@ -63,6 +64,7 @@ import transitions from './shorthands/transitions'
 
 export {
   adjustHue,
+  align,
   animation,
   backgroundImages,
   backgrounds,
--- a/src/shorthands/align.js
+++ b/src/shorthands/align.js
@@ -1,0 +1,164 @@
+// @flow
+
+/**
+ * CSS to represent object in any position.
+ *
+ * @example
+ * // Styles as object usage
+ *
+ * const styles = {
+ *   ...align('right', '-10px')
+ * }
+ *
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   ${align('right', '-10px')}
+ *
+ *
+ * // CSS as JS Output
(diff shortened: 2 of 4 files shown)


bhough:
@tolgaergin Thank you for the PR. While generally I like the idea of an easy way to align items, how does this play with the general move to flexbox for this type of problem. Support is pretty solid now: http://caniuse.com/#feat=flexbox. My initial thought is this is something we'd end up deprecating pretty quickly or have to rework to use flexbox. Any ideas on how you could implement this using flexbox?

tolgaergin:
thanks Brian. I actually thought that the align shorthand might be useful for positioning absolute elements that needs z-index rather than positioning normal elements for layout purpose. 

For example:
1) You can look at the pink notification badge on the user picture on top right. 
2) Grey and pink Badges on top left of the cards  

position align

nikgraf:
While I like the idea I have two concerns:

- This could lead to misunderstandings developers using position absolute more while flexbox might be appropriate.
- Size of the Polished bundle

@bhough @mxstbr there are a couple of useful patterns like this one. What about we encourage people to create micro-libraries and we add list to the readme where we link to those?
@tolgaergin would you be up for that?

tolgaergin:
I'd be happy @nikgraf I thought that you didn't like me :) 
I also like the name that you suggested.

nikgraf:
@tolgaergin cool,can you ping us once you published it as a lib?
why did you think that I didn't like you? was not my intention to give you that feeling

bhough:
Closing this PR, feel free to submit a PR adding it to the readme when it is ready @tolgaergin

PR #106, opened 2017-03-11
zIndex helper

tolgaergin:
Let's imagine, our website have modal lightbox and backdrop div for the modal and also we have a fixed header and a button at the bottom right corner to scroll to top. When we open the modal, we want backdrop div appears on top on the body elements and our model content should appear on top of the backDrop layer. 

We can keep them in an array like above. 
```javascript
const zArray = [
    'modal',
    'backDrop',
    'header',
    'goTop',
    'base',
  ]
```

With polished zIndex helper, The first element of the array which is modal will get the highest z-index value ```z-index: 4``` 

We can write it like that in styled-components
```javascript
const Modal = styled.div `
  position: absolute;
  background-color: #fff;
  ${zIndex('modal', zArray)};
`;
```

And it will generate this css
```css
.modal {
  position: absolute;
  background-color: #fff;
  z-index: 4
}
```

We can also define our own z-index values but we need to keep them in an object instead of an array.

```javascript
const zObject = {
    modal: 1000
    backDrop: 900,
    header: 500,
    goTop: 100,
    base: 1,
  }
```

We need to pass zObject to zIndex helper

```javascript
const Modal = styled.div `
  ...
  ${zIndex('modal', zObject)};
`;
```

We will able to see our modal z-index value that we defined in zObject
```css
.modal {
  ...
  z-index: 1000;
}
```

zIndex helper expects 3 parameters.

**First parameter:** layer name in our stack _(string)_
**Second parameter:** layer Array or layer Object _(array, object)_
**Third parameter:** base value for when calculating z-index. it will be added to reversed index value of the key in our stack if we pass an array _(number)_

```javascript
const Modal = styled.div `
  ...
  ${zIndex('modal', zArray, 50)};
`;
```

```css
.modal {
  ...
  z-index: 54;
}
```

### DIFF
--- a/src/helpers/test/__snapshots__/zIndex.test.js.snap
+++ b/src/helpers/test/__snapshots__/zIndex.test.js.snap
@@ -1,0 +1,24 @@
+exports[`stripUnit should properly add rules when block has existing rules 1`] = `
+Object {
+  "background": "red",
+  "z-index": 0,
+}
+`;
+
+exports[`stripUnit should return correct z-index value in array 1`] = `
+Object {
+  "z-index": 4,
+}
+`;
+
+exports[`stripUnit should return correct z-index value in array with base value 1`] = `
+Object {
+  "z-index": 104,
+}
+`;
+
+exports[`stripUnit should return correct z-index value in object 1`] = `
+Object {
+  "z-index": 23,
+}
+`;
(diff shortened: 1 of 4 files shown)


nikgraf:
While I really like the idea doesn't z-index also depend on the parents z-index and eventually this could become misleading? A tree structure might be more helpful?!?

bhough:
Closing this @tolgaergin. Feel free to re-submit taking into account a solution for the parent problem @nikgraf raised.

PR #116, opened 2017-03-28
Specify files to publish

GertjanReynaert:
This way not all folders and files get published. There is no need for the docs to be in the npm package.

The main reason why I submit this pr is because currently the `.babelrc` file is also published, and causing mayhem in my React Native project.

I'm not sure if the files I specified are sufficient, but it can be a start to cleaner publishes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,11 @@
     "inline-styles",
     "react"
   ],
+  "files": [
+    "package.json",
+    "lib",
+    "flow-typed"
+  ],
   "bugs": {
     "url": "https://github.com/styled-components/polished/issues"
   },


bhough:
Thanks for the PR @GertjanReynaert. We definitely want to clean that up, ASAP.

package.json
```diff
@@ -43,6 +43,11 @@
     "inline-styles",
     "react"
   ],
+  "files": [
+    "package.json",
+    "lib",
```
pablo-coco: "lib" should be "src"

mxstbr: Nope, this is incorrect, it should actually be `lib`.

mxstbr:
Rather than try to figure out which files people need I'd exclude `.babelrc` to fix this bug for now. We can dig into this down the line!

package.json
pablo-coco: I forked the project, installed it through node, but there wasn't any lib folder. The lib folder is build from the src folder, since that folder is not included, it cant build the lib folder. Am I missing something?

mxstbr: You don't need to build the lib folder, that's done before each release so it's on npm. See https://unpkg.com/polished@1.0.0/lib/

mxstbr:
The bug here was fixed in #133, will close this for now. Since I'm not really sure what exactly we want to publish and don't have the head to figure out the pros and cons right now I'll close this PR, but feel free to open an issue and discuss if you feel strongly that we should be whitelisting instead!

GertjanReynaert:
@mxstbr I don't really have an opinion on whitelisting vs ignoring, but I do think that the docs shouldn't be published in the npm package.

mxstbr:
> I do think that the docs shouldn't be published in the npm package.

This is exactly what I want to discuss: why not? I want to look up docs when working from a plane, having them in `node_modules/polished/docs/` is super neat!

Issue #117, opened 2017-03-28
homepage example contains syntax error


rattrayalex:
I pasted this into the console on the homepage: 

```
const styles = {
  color: lighten(0.2, '#000'),
  font-size: modularScale(1),
  [hiDPI(1.5)]: {
    font-size: modularScale(1.25)
  }
}
```
And, unsurprisingly, got
```Uncaught SyntaxError: Unexpected token -```

I assume it needs to be `"font-size": `?

mxstbr:
Yes, good call! PR welcome.

rattrayalex:
Sure, forthcoming

nikgraf:
thanks @rattrayalex

PR #118, opened 2017-03-28
Fix homepage examples

rattrayalex:
addresses https://github.com/styled-components/polished/issues/117

When I downloaded and ran `yarn` it generated a diff; that was my first commit. 

Otherwise, this contains two changes: wrapping `font-size` in `"` on the homepage example, and exporting all the `polished` methods as globals on the homepage, so that just calling `lighten(x)` etc actually works.


docs-theme/assets/script.js
```diff
@@ -1,2 +1,3 @@
+Object.assign(window, polished)
```
mxstbr: This is awesome!

PR #125, opened 2017-03-28
Fix Examples of `opacify`

andrewburgess:
Updated example of `opacify` to show correct outputs, since they
seemed to be copied from `transparentize`

Not sure if I should include the generated docs as well in the commit


src/color/opacify.js
```diff
@@ -12,23 +12,23 @@ import curry from '../internalHelpers/_curry'
  * // Styles as object usage
  * const styles = {
  *   background: opacify(0.1, '#fff');
- *   background: opacify(0.2, 'hsl(0, 0%, 100%)'),
- *   background: opacify(0.5, 'rgba(255, 0, 0, 0.8)'),
+ *   background: opacify(0.2, 'hsl(0, 0%, 20%)'),
```
nikgraf: can you change it to `opacify(0.2, 'hsla(0, 0%, 100%, 0.5)')`
results in: rgba(255,255,255,0.7)

nikgraf:
@andrewburgess thanks for looking into it!

andrewburgess:
@nikgraf updated the example for `hlsa`

PR #127, opened 2017-03-29
docs(placeholder): reflect the usage

evenchange4:
According to the [conversation](https://github.com/styled-components/polished/pull/41#discussion_r91830362), I think the `mixins` do not work with `css` of SC.


bhough:
@mxstbr this appears to be the case, but wanted to confer with you first.

PR #130, opened 2017-03-29
Update README.md

### LINKED ISSUE #128: Clarify use of object rest spread operator in docs?
First of fall: Awesome library! Really cool to throw this at the guys claiming that they "live and die' by Sass functions!

I have suggestion/comment regarding the docs: The examples make extensive use of the object rest spread operator, which is still a proposal, and you may save a lot of developers some time by indicating that they need to use [babel-plugin-transform-object-rest-spread](https://www.npmjs.com/package/babel-plugin-transform-object-rest-spread), in order to leverage the new syntax.

Happy to make a PR to the docs if this is indeed something that should be in there!

jonasws:
Add some instructions on how to enable the object-rest-spread syntax.

Fixes #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,10 @@ Want to write styles in JavaScript, but also want Sass-style helper functions an
 
 You can optionally also use [`babel-plugin-polished`](https://github.com/styled-components/babel-plugin-polished) to compile the static function calls out and remove the (already tiny) runtime performance impact of using `✨ polished`.
 
-### Object rest spread syntax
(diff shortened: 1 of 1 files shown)


bhough:
Thanks for the PR @jonasws. One small change and rebase on master and we should be good to go.

jonasws:
I didn't realize the syntax was no longer experimental, I knew it was not in the ES2015 preset. Will make the change sometime today hopefully!

PR #131, opened 2017-03-29
Fixed typo in documentation

mob8607:
https://github.com/styled-components/styled-components/blob/master/docs/api.md#injectglobal

The method name should be injectGlobal instead of injectGlobals or am I missing something?


bhough:
Thanks @mob8607, good catch.

PR #136, opened 2017-03-31
WP - Error and Warning Handling

bhough:


nikgraf:
@bhough do we want to tackle this? It's not urgent. Could be done in a hacking session when you visit Vienna or Max and I travel to SF :)

bhough:
Definitely do! I think I'm finally happy with the API now that I've found a way to suss out the file/module info so it does not have to be passed in every call. I also have it properly pulling out the error handling in the production build.  The next thing I'm trying to work through is how we should design the different message types. The four types of message I think we will need to support:

1. Warning - Something incorrect the user is doing that doesn't impact how this modules works (i.e. passing extra parameters).
2. Error that would cause a CSS Issue - Something the user is doing that will not cause the module to fail on the JS side but would produce invalid CSS.
3. Error that would cause a JS issue - Something the user is doing that will cause the module to fail on the JS side.
4. Deprecation warning - Likely something that would be added in a separate ticket, but the ability to warn a user when a module they are leverage will be deprecated in a later release.

As part of this, also need to determine an order of priority for checking for common errors across modules. Such as `number of arguments > argument order > individual argument type > individual argument value`.

Always open for a hacking sessions. I am actually in Basel, Switzerland until the end of the year, so much closer than SF. :)

PR #142, opened 2017-04-05
chore(color): Allow whitespace in rgb and hsl values.

10xjs:
As per the spec: https://drafts.csswg.org/css-color-3/#rgb-color.

> White space characters are allowed around the numerical values.


bhough:
This looks good to me, but I'd like @nikgraf to give the final sign off since he is more familiar with this mixin.

nikgraf:
yep, looks good. thanks @10xjs 🙌

PR #146, opened 2017-04-11
Typo

aymericbouzy:
Or else it's a syntax I don't know about ...


bhough:
@aymericbouzy That is the ESNext Spread Syntax. You can read more about it here: https://github.com/styled-components/polished#object-spread-properties

aymericbouzy:
@bhough spread in a template literal? I insist I don't know about this syntax, although I do know about the spread operator. http://stackoverflow.com/a/42643865/3995091

Issue #147, opened 2017-04-11
steps in modularScale() is not calculated correctly


boxfish:
Given default base (`1em`) and ratio (`perfectFourth`), modularScale(2) should return `1.77689em`, but currently it is returning `2.666em`. 

It seems the current output is calculated by `base * ratio * steps`, but it should be `base * Math.pow(ratio, steps)`

yeskunall:
Looks like [this](https://github.com/styled-components/polished/blob/master/src/helpers/modularScale.js#L82) confirms what @boxfish reported. Based on [modularscale.com](http://www.modularscale.com/?1&em&1.333), modular scale is calculated as `base * Math.pow(ratio, steps)`.

mxstbr:
Yes please, PR very welcome!

brikou:
@mxstbr I just made a PR for this one and another bug when steps is 0, can you review please #149 

Thank you for your work :+1:

bhough:
Closed by #149. Thanks @brikou!

PR #149, opened 2017-04-12
fix(helpers): Fix modularScale computation, allow steps to be zero

brikou:
Fix issue #147 and also allow steps to be set to zero


brikou:
Everything is green, ready for review

brikou:
@bhough using Math.pow is not possible as it raise an eslint error.

bhough:
@brikou ok, why that shouldn't be a blocker to using `Math.pow`, looking over it this weekend, I don't think we will get numbers big enough to trigger the Chrome bug. Will keep an eye on it in case it gets used in a way I'm not anticipating that raises the issue.

PR #154, opened 2017-04-22
feat(typings): Add typescript declarations

mvestergaard:


bhough:
Thanks for the PR @mvestergaard, definitely something we want to add. Will give this a look this week.

mvestergaard:
@mxstbr I don't think there's any good solution for that sort of thing. The easiest would probably be to have a checklist template for PRs

mxstbr:
> to have a checklist template for PRs

As we all know, humans can't do any mistakes when checklists are around :wink: I'm a bit surprised, is there really no way to make sure we have 100% TS coverage?

mvestergaard:
Well there was if the project was written in TS, but unfortunately there's a divergence between flow and TS, that's not easily fixed.

I've been googling a bit for a way to convert flow types to TS, but can only find things that do it the other way. Converted/generated types could also lead to worse quality.

staltz:
https://twitter.com/mxstbr/status/856861482067386369

> Can tslint lint our entire codebase for missing typings? I'd hate for the TS types to get out of sync with the code because folks submit PRs and we forget to think about it.

So basically no, because TypeScript compiler `tsc` doesn't look into JS files (like Flow does), `tsc` just looks at .ts and .d.ts to generate JS. Nowadays there is the `tsc` option `--allowJs` which will take JS files as input and output it as JS, which AFAIK doesn't do any type checking, only performs module transformation like ES6 module to commonJS.

That said, there are tools like https://github.com/webbiesdk/tstools, which you could leverage, but I haven't seen it used for that purpose, so far.

- - -

The PR itself looks decent and correct. I haven't used `tslint` for linting `.d.ts` files, only `.ts` files, so I'm not sure if that would work as expected. You need to run the new scripts and check if it passes. Note that those scripts aren't currently called from other scripts, that may be a problem.

- - -

PS: `polished` looks great. Nice to hear about this project. :)

mvestergaard:
Regarding this bit

> Note that those scripts aren't currently called from other scripts, that may be a problem.

I pretty much copied the setup from the main styled-components repo, which I presume was the preferred setup.

nikgraf:
@bhough @mxstbr any objections merging this one?

bhough:
@nikgraf generally speaking this all looks good to me, and people who know more about TS than I have signed off on the code. Up to @mxstbr given his concerns above.

sebald:
What's the status of this PR? Can I help somehow to get this merged? :)

qpowell:
@mxstbr 👋

package.json
```diff
@@ -24,7 +25,9 @@
     "lint": "eslint src",
     "flow": "flow check",
     "docs": "pushstate-server docs",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "tslint": "tslint typings/*.ts",
+    "typescript": "tsc ./typings/polished-test.ts --noEmit --target es6 --module es2015 --moduleResolution node"
```
luke-john: There's a range of things in typescript that can go wrong when using definitions when generating definitions.

Testing this adds a bit of extra complexity but is worthwhile imo.

https://github.com/paypal/glamorous/pull/124

bhough:
Closing this in favor of #215 which is using https://github.com/ForbesLindesay/tsgen to automatically generate TS definitions.

Issue #153, opened 2017-04-18
Support for named css colors


suchipi:
It'd be nice if you could use helpers with named CSS colors, eg "white", "salmon", etc. Polished would need to have a map of names to "usable" colors. It could maintain its own list of colors, or use a package like `css-color-names`.

bhough:
@suchipi Thank you for the suggestion. I'm assuming this would be so you could use named colors with the color mixins? Ideally what would these be mapped to? I'm assuming the hex values?

suchipi:
Yeah, I was thinking the hex values; to be used with darken, desaturate, [etc](https://polished.js.org/docs/#color).

bhough:
Definitely something we'd consider doing. The concern I have that would need addressed is to how to add that functionality to all the color mixins without bloating the size of those modules.

suchipi:
What kind of size goals does the project have? It looks like my install of `polished.min.js` is 24KB.

Here's one example of a name-to-hex map (not sure if it's complete), that's 3.75KB uncompressed: https://github.com/bahamas10/css-color-names/blob/master/css-color-names.json

Something like this could be shrunk by removing whitespace, duplicated `#` characters, and converting 6-character hex codes to 3 characters where possible.

It's also possible to offload this work to the browser via `getComputedStyle`, but this would not work in node, and likely not work in jsdom. I was under the impression that it was a goal of this project to work in node/jsdom as well. Is that correct?

If there was a way to make this theoretical part of the codebase tree-shakable if unused, would that change the size requirements at all?

What if a "browser" build relied on `getComputedStyle`, but a "node" build relied on a map in the bundle?

Another approach would be to try to do this work with a babel plugin. However, that would not cover dynamic cases (it could only transform string literals).

bhough:
Right now we don't have a size goal for the entire project as we want users to only import the modules they need. We also are aiming for this to be dependency-free.

Yes we definitely want it to work across the board, so `getComputedStyle` is not an option, and I don't want to fork the builds based on the target for one feature.

It would be tree-shakable as far as it would only be included if a color module that used it was imported. But without dynamic imports in ES6, there isn't really a clean way to drop the named color list on its own if it isn't used.

To be clear, I think we can get this to an acceptable state, just want to make sure we consider all the implications. That list you linked is complete, so that gives us a baseline. Would love to get a PR started and we can start iterating on the approach.

nikgraf:
Really nice considerations!

bhough:
[...]

PR #155, opened 2017-04-23
[WIP] feat(color): Named color support for color mixins.

bhough:
Adds support for CSS named colors in all color mixins. Also fixes shade and tint docs.

Addresses request in #153


bhough:
@mxstbr @nikgraf @scuhipi I'm of two minds on this one:

It is really nice to just be able to use named colors in all our color methods without any extra effort. But will they be used frequently enough to warrant the ~4kb overhead on every color mixin? Would it be better to just expose `nameToHex` as a helper for the times someone wants to pass a named color into a mixin?

suchipi:
You could save some size and duplication by changing parseToRgb, parseToHsl, etc to all call nameToHex internally. Then you wouldn't need to repeat the call in each helper (though it's a little more coupled organizationally).

I personally prefer the helpers handling named colors out of the box because that behavior matches what sass and stylus helpers do. When I ran into this, it was surprising to me to find this wasn't supported by polished out of the box,  because someone had described polished to me as "common sass helpers for js". Exporting nameToHex (instead of calling it internally) would provide the tools necessary to handle named colors, but handling it internally provides a smoother developer experience for those who need it, since they don't have to go through the roundabout of realizing they need to handle this case and adding the helper (if they do). So to me it seems like a question of smoother DX ("pit of success") or smaller bundle size.

src/color/opacify.js
```diff
@@ -32,7 +33,8 @@ import curry from '../internalHelpers/_curry'
  * }
  */
 function opacify(amount: number, color: string) {
-  const parsedColor = parseToRgb(color)
+  const normalizedColor = nameToHex(color)
```
nikgraf: @bhough I think we can drop `nameToHex ` here since it's already done by default in `parseToRgb`. Same goes for all the other functions including the ones using `parseToHsl`

bhough:
Good catch @nikgraf. I'll get that taken care of at some point before the weekend.

bhough:
@nikgraf this is ready to re-review.

src/color/grayscale.js
```diff
@@ -26,8 +26,9 @@ import toColorString from './toColorString'
  * }
  */
 function grayscale(color: string): string {
+  const hslColor = parseToHsl(color)
   return toColorString({
-    ...parseToHsl(color),
```
nikgraf: @bhough one nitpick … can you revert that change? No need to assign the hslColor here

src/color/parseToHsl.js
```diff
@@ -16,9 +17,10 @@ import type { HslColor, HslaColor } from '../types/color'
  * const color2 = 'hsla(210, 10%, 40%, 0.75)';
  */
 function parseToHsl(color: string): HslColor | HslaColor {
+  const normalizedColor = nameToHex(color)
   // Note: At a later stage we can optimize this function as right now a hsl
   // color would be parsed converted to rgb values and converted back to hsl.
-  return rgbToHsl(parseToRgb(color))
+  return rgbToHsl(parseToRgb(normalizedColor))
```
nikgraf: no need for an assignment here, best to just inline it

src/color/setSaturation.js
```diff
@@ -28,8 +28,9 @@ import curry from '../internalHelpers/_curry'
  * }
  */
 function setSaturation(saturation: number, color: string): string {
+  const hslColor = parseToHsl(color)
   return toColorString({
-    ...parseToHsl(color),
+    ...hslColor,
```
nikgraf: same here, inlining it should save us a couple bites again

bhough: ok I originally did this because we were doing it two different ways across color modules, I'll have to change this back as well as the update the ones where we were doing an assignment.

bhough:
@nikgraf ok third time is a charm. I actually cleaned it up some more, found a few more things that eventually call `parseToRGB` so was able to re move a few more redundant `nameToHex` calls.

PR #161, opened 2017-05-02
fix(cov): avoid publishing coverage results

chambo-e:


nikgraf:
Thanks for the fix @chambo-e 👏

chambo-e:
You're welcome :slightly_smiling_face:

Issue #160, opened 2017-05-01
Triangle mixin needs completed


DJTB:
The docstring for Triangle currently ends up in the built docs under Types > PointingDirection

I went for a quick PR but ran into flow annotation issues due to lack of familiarity with it.

Seems all that's needed is to add `"triangle`" under Mixins in `.documentation.js` and to correct the docstring placement from above `PointingDirection` to above `triangle` in `src/mixins/triangle.js` (and whatever flow wants to be happy ^^).

Moreover, triangle might need to be switched from const arrow => to plain function declaration to match the main export style in the other src files.

DJTB:
Oh, I just realised it's not exported in the main package either (attempted to use es6 import).
Not sure if that's an oversight or intentional!

bhough:
Thanks @DJTB. It's an oversight, we will take a look at getting that corrected this week.

bhough:
@nikgraf @mxstbr looks like this needs a bit of re-working to match our other APIs, documentation is incorrect, and types aren't working exactly right. Will spend some time working on this, this week. Not a bug per se, as it is not usable, but should get it added.

bhough:
Closed by #162

PR #162, opened 2017-05-03
fix(triangle): Complete triangle mixin

bhough:
Finishes the triangle mixin with an updated API to be more in line with Bourbon's implementation.

Also updates the associated docs and properly exports the module.

#160


nikgraf:
@mxstbr can you review this one?

PR #163, opened 2017-05-03
feat(getLuminance): Add getLuminance function

### LINKED ISSUE #148: Color Luminance utility functions
Would you be interested in adding a luminance helper to this library? I'm using this right now which pulls from the knowledge of other color libraries:

```js
// http://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef
function getLuminance(color) {
  const rgbColor = parseToRgb(color);
  const [r, g, b] = Object.keys(rgbColor).map(key => {
    const channel = rgbColor[key] / 255;
    return channel <= 0.03928
      ? channel / 12.92
      : Math.pow((channel + 0.055) / 1.055, 2.4);
  });
  return 0.2126 * r + 0.7152 * g + 0.0722 * b;
}
```

ported this from [ChromaJS](https://github.com/gka/chroma.js/blob/d2c6d917df4ba2b87d8a740de116a0656bcbdfd5/src/io/luminance.coffee) because I didn't want to import the whole lib, not sure if you want to put it in here though
```js
const EPS = 1e-7;

function setLuminance(amount, color) {
  const { red, green, blue, alpha = 1 } = parseToRgb(color);
  let rgb;
  if (amount === 0) {
    rgb = rgba(0, 0, 0, alpha);
  } else if (amount === 1) {
    rgb = rgba(255, 255, 255, alpha);
  } else {
    let maxIteration = 20;
    const test = (color1, color2) => {
      const mixed = mix(0.5, color1, color2);
      const mixedLuminance = getLuminance(mixed);
      if (Math.abs(amount - mixedLuminance) < EPS || !maxIteration--) {
        return mixed;
      }
      if (mixedLuminance > amount) {
        return test(color1, mixed);
      }
      return test(mixed, color2);
    };
    rgb = getLuminance(color) > amount
      ? test('#000', color)
      : test(color, '#fff');
  }
  return rgb;
}
```

Could even add some simple `isDark` and `isLight` functions with this addition. I can put together a PR if you are interested, just let me know :)

jcquinlan:
Closes #148 

At least partially, as it allows the user to get the value, but doesn't have an accompanying setter to change the luminance. It seemed like this was already being handled by other methods.

I feel like to really make this useful, another method would need to allow a user to compare the contrast ratio of two colors and return a boolean stating whether or not it adheres to the [W3C specs](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html#contrast-ratiodef).

### DIFF
--- a/src/color/getLuminance.js
+++ b/src/color/getLuminance.js
@@ -1,0 +1,41 @@
+// @flow
+
+import parseToRgb from './parseToRgb'
+import curry from '../internalHelpers/_curry'
+
+/**
+ * Returns a number representing the luminance of a color.
+ *
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   background: getLuminance('#CCCD64') >= getLuminance('#0000ff') ? '#CCCD64' : '#0000ff',
+ *   background: getLuminance('rgba(58, 133, 255, 1)') >= getLuminance('rgba(255, 57, 149, 1)') ?
+ *                             'rgba(58, 133, 255, 1)' :
+ *                             'rgba(255, 57, 149, 1)',
+ * }
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   background: ${getLuminance('#CCCD64') >= getLuminance('#0000ff') ? '#CCCD64' : '#0000ff'};
+ *   background: ${getLuminance('rgba(58, 133, 255, 1)') >= getLuminance('rgba(255, 57, 149, 1)') ?
+ *                             'rgba(58, 133, 255, 1)' :
+ *                             'rgba(255, 57, 149, 1)'};
+ *
+ *
+ */
+function getLuminance(color: string): number {
+  const rgbColor = parseToRgb(color);
+  const [r, g, b] = Object.keys(rgbColor).map(key => {
+    const channel = rgbColor[key] / 255;
+    return channel <= 0.03928
+      ? channel / 12.92
+      : Math.pow((channel + 0.055) / 1.055, 2.4);
+  });
+  return 0.2126 * r + 0.7152 * g + 0.0722 * b;
+}
+
+
+// Don’t inline this variable into export because Rollup will remove the /*#__PURE__*/ comment
+const curriedGetLuminance = /*#__PURE__*/ curry(getLuminance) // eslint-disable-line spaced-comment
+export default curriedGetLuminance
--- a/src/color/getLuminance.js
+++ b/src/color/getLuminance.js
@@ -1,6 +1,5 @@
 // @flow
 import parseToRgb from './parseToRgb'
-import curry from '../internalHelpers/_curry'
 
 /**
  * Returns a number (float) representing the luminance of a color.
@@ -39,6 +38,4 @@ function getLuminance(color: string): number {
   return 0.2126 * r + 0.7152 * g + 0.0722 * b
 }
 
-// Don’t inline this variable into export because Rollup will remove the /*#__PURE__*/ comment
-const curriedGetLuminance = /*#__PURE__*/ curry(getLuminance) // eslint-disable-line spaced-comment
-export default curriedGetLuminance
+export default getLuminance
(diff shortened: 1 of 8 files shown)


bhough:
Thanks for the PR @jcquinlan. I definitely think we should at least add some sort of guard type functionality http://lesscss.org/features/#mixin-guards-feature that allows us to check the lightness or darkness of a color. There are also things we can do related to contrast here that would also be useful: https://codepen.io/giana/project/full/ZWbGzD

But I agree, `getLuminance` isn't necessarily useful on its own.

souporserious:
Awesome to see this in a PR! I was thinking about it, and do you think it should be called `getLightness` to be in line with `setLightness`?

jcquinlan:
@souporserious Hey sorry I ripped off your idea! I was bored at a conference and wanted to do some coding, I hope you don't mind!

As per the suggestion of @bhough it might be a good idea to use this as a stepping stone to creating some of the guards and more full fleshed-out brightness tools that exist elsewhere.

souporserious:
Oh no, not at all! I'm glad someone found the time to implement it 🙂

wtgtybhertgeghgtwtg:
Other than the conflicting files, is there anything in particular that is holding this up?

bhough:
@wtgtybhertgeghgtwtg it needs several things:

1. Conflicts resolved
2. Proper eslint comments per: https://github.com/styled-components/polished/pull/204
3. Ideally, I believe it was intended to work in conjunction with: https://github.com/styled-components/polished/pull/217/files
4. Would also need tests for the other color spaces

src/color/test/getLuminance.test.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import getLuminance from '../getLuminance'
+
+describe('getLuminance', () => {
```
bhough: Would like to see the tests for all color spaces if possible (rgb, hsl, hsla, and a named CSS color) if we could.

src/color/getLuminance.js
```diff
@@ -0,0 +1,38 @@
...
+  const [r, g, b] = Object.keys(rgbColor).map(key => {
+    const channel = rgbColor[key] / 255
+    return channel <= 0.03928
+      ? channel / 12.92
+      : ((channel + 0.055) / 1.055) ** 2.4
+  })
+  return (0.2126 * r) + (0.7152 * g) + (0.0722 * b)
+}
+
+export default curry(getLuminance)
```
bhough: This will need to be updated based on #204 

```diff
- export default curry(getLuminance)
+ // Don’t inline this variable into export because Rollup will remove the /*#__PURE__*/ comment
+ const curriedGetLuminance = /*#__PURE__*/ curry(getLuminance) // eslint-disable-line spaced-comment
+ export default curriedGetLuminance
```

src/color/getLuminance.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+import parseToRgb from './parseToRgb'
+import curry from '../internalHelpers/_curry'
+
+/**
+ * Returns a number representing the luminance of a color.
```
bhough: Maybe clarify that this is a percentage (float)?

README.md
```diff
@@ -146,6 +146,7 @@ In the documentation you will see examples using [object spread properties](http
     <li><a href="http://polished.js.org/docs/#rgbacolor">RgbaColor</a></li>
     <li><a href="http://polished.js.org/docs/#rgbcolor">RgbColor</a></li>
     <li><a href="http://polished.js.org/docs/#timingfunction">TimingFunction</a></li>
+    <li><a href="http://polished.js.org/docs/#getluminance">getLuminance</a></li>
```
bhough: This is getting put in the wrong spot, due to it missing from `documentation.json`. If you add it, this *should* get inserted into the proper place in the readme and the documentation.

bhough:
@jcquinlan I was just starting to look at this myself, so I already merged the rebase with master into your fork 😎

PR #111, opened 2017-03-23
Error/Warning Handling

bhough:
Handler for pretty printed warnings and errors in console in development mode.

PR #166, opened 2017-05-07
Permit number inputs including allowing zero as a valid value.

dlwalsh:
This change allows `0` as a valid value for the `directionalProperty` helper. Previously a zero value would be ignored and left unset.

In practical terms, it means you can type `position(0)` whereas before you had to type `position('0')`.

### DIFF
--- a/src/mixins/test/ellipsis.test.js
+++ b/src/mixins/test/ellipsis.test.js
@@ -6,6 +6,10 @@ describe('ellipsis', () => {
     expect({ ...ellipsis('300px') }).toMatchSnapshot()
   })
 
+  it('should permit bare number', () => {
+    expect({ ...ellipsis(300) }).toMatchSnapshot()
+  })
+
   it('should properly add rules when block has existing rules', () => {
     expect({
       background: 'red',
--- a/src/shorthands/size.js
+++ b/src/shorthands/size.js
@@ -21,7 +21,7 @@
  * }
  */
 
-function size(height: string, width: string = height) {
+function size(height: string | number, width: string | number = height) {
   return {
     height,
     width,
--- a/src/shorthands/test/size.test.js
+++ b/src/shorthands/test/size.test.js
@@ -16,4 +16,12 @@ describe('size', () => {
   it('should set height and width to the same value when only one parameter is passed', () => {
     expect({ ...size('300px') }).toMatchSnapshot()
   })
+
+  it('should set height and width with unitless values', () => {
+    expect({ ...size(300, 250) }).toMatchSnapshot()
+  })
+
+  it('should set height and width with unitless zero', () => {
+    expect({ ...size(0) }).toMatchSnapshot()
+  })
 })
--- a/src/helpers/directionalProperty.js
+++ b/src/helpers/directionalProperty.js
@@ -8,10 +8,10 @@ function generateProperty(property: string, position: string) {
   return splitPropertyName.join('-')
 }
 
-function generateStyles(property: string, valuesWithDefaults: Array<?string>) {
+function generateStyles(property: string, valuesWithDefaults: Array<?string | ?number>) {
   const styles = {}
   for (let i = 0; i < valuesWithDefaults.length; i += 1) {
-    if (valuesWithDefaults[i]) {
+    if (valuesWithDefaults[i] || valuesWithDefaults[i] === 0) {
       styles[generateProperty(property, positionMap[i])] = valuesWithDefaults[i]
     }
   }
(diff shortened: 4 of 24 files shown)


bhough:
Thanks for the PR @dlwalsh. While I don't have any immediate issue with this change, it would leave us with an inconsistent API across modules. Off the top of my head, `borderRadius`, `borderWidth`, `margin`, `padding`, `position`, and `size` would also need to be updated to allow a non-string zero value as well.

dlwalsh:
Updated `borderRadius`, `borderWidth`, `margin`, `padding`, `position` and `size` to permit numeric inputs, including new tests and snapshots.

For the most part, the only source changes were updating flow definitions. The one exception is `borderRadius` where the exception throw logic has changed.

bhough:
@dlwalsh if you want to update this PR to be against V2, I'd like to include it in that release.

dlwalsh:
Rebased on the version-2 branch. I've also changed `ellipsis` to allow a bare number.

src/shorthands/size.js
```diff
@@ -21,7 +21,7 @@
  * }
  */
 
-function size(height: string, width: string = height) {
+function size(height: string | number, width: string | number = height) {
```
bhough: Unitless height/width is invalid unless 0, so we should put the same `=== 0` check here as in the other modules.

src/shorthands/test/size.test.js
```diff
@@ -16,4 +16,12 @@ describe('size', () => {
   it('should set height and width to the same value when only one parameter is passed', () => {
     expect({ ...size('300px') }).toMatchSnapshot()
   })
+
+  it('should set height and width with unitless values', () => {
```
bhough: See above comment about unitless height/width

src/mixins/test/ellipsis.test.js
```diff
@@ -6,6 +6,10 @@ describe('ellipsis', () => {
     expect({ ...ellipsis('300px') }).toMatchSnapshot()
   })
 
+  it('should permit bare number', () => {
```
bhough: Unitless width is invalid so we should not allow this.

dlwalsh:
Actually @bhough, permitting numbers is the real intent of the change. Many css-in-js libs will add the units automatically, e.g. glamorous, jss. This is a convenience thing. For instance, you can perform mathematical operations on raw numbers.

I know I specifically focused on `0` in the original post. But `0` was a special case because the existing logic specifically disallowed falsy values. The library has always passed through (non-zero) numbers. I'm simply bringing the static type definitions into line with how I (and presumably others) use the library.

IMHO polished need not necessarily generate valid CSS. It just needs to produce what css-in-js libs can turn into valid CSS. Perhaps this is a topic for further discussion?

bhough:
Thanks for raising this, there is definitely more discussion/thought needed here.

Styled-components does not add values automatically. Related discussion as it pertains to the impact that could have on React Native: https://github.com/styled-components/styled-components/issues/835. It is aiming to solve both React and Native styling with one library. However, glamorous for example, has a separate Native implementation instead.

I'd rather move forward with for the the reasons outlined in that link than because of libraries automatically adding units. It seems forcing `px` actually could cause issues for React Native development in libraries that force units. We definitely don't want to add to that problem and what to be as usable as possible for Native as well as web.

The question will become how to validate this. Inputs could fall under one of three categories. Valid CSS, Valid for Native styles (but not necessarily valid CSS), and invalid for both. Potentially need to add environment detection into #111 so we can decide which rules to apply?

> IMHO polished need not necessarily generate valid CSS. It just needs to produce what css-in-js libs can turn into valid CSS.

This requires that 'what can be turned into valid CSS' is consistent across libraries, which isn't the case right now. If it always produces valid CSS (or React Native styles) than it removes any question of if it will work with current or future libraries.

Thoughts?

dlwalsh:
> This requires that 'what can be turned into valid CSS' is consistent across libraries

This is where I disagree. It's the responsibility of the developer to know what can be fed to the css-in-js lib. If I'm using glamor(ous), I know that I can use numbers or strings, and that numbers are automatically appended with `px`. If I'm using styled-components, I should be aware that I must supply strings.

Basically, I don't see validation being the responsibility of polished. It simply maps the provided arguments.

bhough:
So I want to separate the unitless value discussion from the rest of the debate. As it has some unique issues not present in general validation. 

Allowing unitless values are important due to React Native, and is something I blanked on when making my original comment. Though there still seems to be some debate about the best approach for these cases of universal components, and the best approach right now is to stay as un-opinionated as we can until the dust settles. Which means, allowing numbers or strings at this point as you have in this PR.

> I disagree. It's the responsibility of the developer to know what can be fed to the css-in-js lib.

I look at this differently. I see it as making polished code more portable between libraries. If I transition my 'polished enhanced' styles from one solution to the other it always produces the same result.

> Basically, I don't see validation being the responsibility of polished. It simply maps the provided arguments.

There are layers of validation, so we should be careful saying that validation isn't polished's responsibility, as I believe validation of polished's API *is* it's responsibility and different than validating that, for example, a string is a proper value for CSS property.

These discussions are why #111 has been so much "fun" 😬 . I'll DM you my approach as I don't want to liter this PR with unrelated validation discussion, but want to bounce my approach off of you.

Issue #104, opened 2017-03-09
[V2] rgba function should behave like Sass rgba function


tolgaergin:
The rgba function that you created expects 
```${rgba(255, 205, 100, 0.7)}``` or 
```${rgba({ red: 255, green: 205, blue: 100, alpha: 0.7 })};``` 
when we use it in styled-components. 

I don't know, my approach might be wrong but If I use your rgba function like ```${rgba('#000', 0.1)}``` in order to get the same css ```rgba(0, 0, 0, 0.1)``` when I use rgba function in [Sass,](https://github.com/sass/node-sass/blob/1f5eaff36a7e06126df0902df681aa980250be55/src/libsass/src/functions.cpp#L267) it would be better.

I use this function to get the result that I mentioned.
```javascript
const hexToRgb = (hex) => {
  // http://stackoverflow.com/a/5624139
  var shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
  hex = hex.replace(shorthandRegex, function (m, r, g, b) {
    return r + r + g + g + b + b;
  });

  var result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
  return result ? {
    r: parseInt(result[1], 16),
    g: parseInt(result[2], 16),
    b: parseInt(result[3], 16),
  } : null;
};

const rgba = (hex, alpha) => {
  const color = hexToRgb(hex);
  return `rgba(${color.r}, ${color.g}, ${color.b}, ${alpha})`;
};

export const Header = styled.div `
  box-shadow: 0 0 5px ${rgba('#000', 0.1)};
`;
```

I wonder your ideas about it.

I look forward to use Polished in a real project.
Thanks

nikgraf:
Interesting case. The functionality is already covered by https://polished.js.org/docs/#transparentize

To be honest I'm not sure. I thought this would not be needed, but if this is a very common case for Sass users we might want to consider supporting it.

tolgaergin:
Thanks for your thoughts Nik,

I've been asking many people around me for 2 weeks. the most common built-in functions in sass in my circle like that 1) rgba 2) lighten, darken

Additionally, I've investigate the most [popular front-end libraries](http://cssdb.co/) uses Sass, 
if I compare their usage of rgba(#hex, alpha) or transparentize(alpha, #hex)

| Library | rgba usage | transparentize usage
| --- | --- | --- |
[...]

PR #168, opened 2017-05-10
feat(rgba): Fade color with rgba module

bhough:
rgba now can accept a hex or named CSS color value and an alpha value to fade a color.

Addresses request made in #104


nikgraf:
Still not a fan, `transparentize(0.3, '#fff')` is a much more descriptive API :)

Nevertheless the PR looks good. One open question: Should we allow strings then rgb as well? If not that seems to be pretty inconsistent.

bhough:
@nikgraf I've been kinda torn on this one myself. There are a couple things at play here.

1) it seems like the consensus is that users don't seem to want to do the math and would rather have the alpha value just be a pass in for the alpha level as opposed to the amount you want to transparentize by.

2) because the goal of the project is still to replace sass/less and their libraries, users seem to want rgba to work the same way.

Which still leaves us with the question of, is there a solution that accomplishes these things without making us feel bad about including it in the library?

Can you give me an example of what you mean when you say :

> Should we allow strings then rgb as well?

nikgraf:
Hmm, I can't remember what I meant by strings, but the PR looks good to me.

If we haven't found a better solution after that time I think the best course of action is to merge it :)

bhough:
This is not a breaking change, so I'm fine with merging it.

Issue #150, opened 2017-04-13
RFC change modularScale default unit from `em` to `rem`


brikou:
`em` units are based on the font size of the element they’re used on whereas `rem` units are based on the font size of the html element.

I suggest to update modularScale signature to use `rem` as default.

read more here https://webdesign.tutsplus.com/tutorials/comprehensive-guide-when-to-use-em-vs-rem--cms-23984

WDYT ?

brikou:
EDIT: in fact I think that default base should be simply 1 (as float). This helper shouldn't have anything to do with unit IMO.

brikou:
The more I think about it... the more I feel that I am wrong and `1em` as base unit is perfect... So closing this issue

brunnolou:
@brikou can you explain why you now think `em` is better instead of `rem` in modular scale?

IMHO `modularScale(2)` should always be the same `px` size, and not be determined by the font size of the parent.

PR #179, opened 2017-05-20
Flexible modularScale output unit

nagelflorian:
The output of `modularScale` is currently hard-coded to use `em` for its ouput. In an attempt to avoid this assumption and make the function return other units this PR extends `modularScale` to parse the unit from the base argument.

Related to #150 (RFC change modularScale default unit from `em` to `rem`).

### DIFF
--- a/src/helpers/modularScale.js
+++ b/src/helpers/modularScale.js
@@ -1,5 +1,6 @@
 // @flow
 
+import getUnit from '../internalHelpers/_getUnit'
 import stripUnit from './stripUnit'
 
 const ratioNames = {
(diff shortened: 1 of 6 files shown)


nagelflorian:
Hey @mxstbr, thanks for the feedback. Good idea. I've updated the `getUnit` tests to also include `.05` inputs.

nikgraf:
@bhough should we close this one or fix it ourselves?

nagelflorian:
@nikgraf as I won't have the time to make the necessary updates in the next 1-2 weeks please feel free to close this.

nikgraf:
@nagelflorian 👍

angeloashmore:
Hi, I know this is already closed, but are there any plans to implement this functionality in the future? I rarely have a use for `em`s, but providing an arbitrary unit would be helpful.

PR #181, opened 2017-05-23
chore(types): Add “exact” object types to the color types

Alex4S:


nikgraf:
nice contribution @alex4s

Issue #177, opened 2017-05-18
[V2] [Discussion] Contributor Experience Improvements


bhough:
Some items I'd like to discuss for V2 to help our day to day development workflow:

- [x] Implement prettier
- [x] NPM publish audit (users are indicating we may still be publishing unnecessary files to npm)
- [x] Cleanup Build Scripts (we keep ending up with an out of date bundle checked into the repo)
- [x] Replace ghooks [DEPRECATED] with husky

mxstbr:
>  Remove semantic-release

Why do we want to remove that? I quite like not having to go through and publish each change manually...

bhough:
> Why do we want to remove that? I quite like not having to go through and publish each change manually...

@mxstbr I don't feel super strong about this one. @nikgraf and I were just talking at ReactConf that sometimes it can be a bit cumbersome to work with. For example, we had a few releases get missed due to mistyped commit messages by contributors (that were manually doing the format instead of using commitizen). Also a standing question about how we use it to generate whole point releases. 

I'm not opposed to sticking with it, just wanted to discuss those points.

bhough:
Just a note that I'm still evaluating some of the changes made specifically to the build script. The one challenge with building the docs on every commit is that it makes rebasing a bit of a nightmare as every commit ends up with some sort of conflict you have to account for when rebasing.

nikgraf:
@bhough since all of them are checked, can we close this issue?

bhough:
@nikgraf Yeah for now, I may tweak the precommit script but that can be done separately.

PR #184, opened 2017-05-25
[V2]: Contributor Experience Improvements

bhough:
- Implements prettier
- Adds precommit script to run `prettier` and `eslint --fix` on staged files before commit
- Transitions githook handling from `ghooks` to `husky`
- Adjusts eslint rule for mixed operators to play nice with `prettier` and `--fix`
- Updates `.npmignore` to only publish essential files and folders to reduce installed package size
- Adds `test:watch` script to run jest in watch mode
- Adds `cli` script to run all cli related scripts
- Adds `build:docs` to ensure docs are being rebuilt before commit.

Addresses #177 

***Note: This should get merged into the `version-2` branch first, as all subsequent PRs will need to use this new build setup.***


bhough:
@nikgraf @mxstbr A few points of discussion here:

1. This is a pretty aggressive `.npmignore` it only publishes the pure essentials to use the package. If the user wants to modify it locally they will need to fork the repo. I personally like this approach, but there are certainly differing opinions here.

2. I added `build:docs` to the `precommit` hook whenever a `js` file changes. This does means the docs will rebuild on every commit, which may be a bit heavy handed. However, at least half the time contributors are not remember to rebuild the docs and either miss it altogether or add additional rebuild commits. This causes somewhat confusing PRs when a later commit rebuilds the docs and shows changes from previous commits. This would solve that.

3. I turned off the `eslint` error for `no-mixed-operators` as `prettier` breaks this and `eslint --fix` is not capable of restoring the previous formatting. There is no `prettier` flag for turning this off. Seemed like a minor thing, but wanted to point it out.

package.json
```diff
@@ -30,10 +30,16 @@
...
+    "src/**/*.js": [
       "prettier --no-semi --single-quote --trailing-comma all --write",
       "eslint src --fix",
       "build:docs",
+      "git add docs",
+      "git add"
+    ],
+    "src/**/*.*.js": [
+      "prettier --no-semi --single-quote --trailing-comma all --write",
+      "eslint src --fix",
```
luftywiranda13: I think passing path isn't necessary here.. because lint-staged runs against staged files and src dir is already passed as filter
<br />
screen shot 2017-05-25 at 23 02 59
*from **lint-staged** docs*

luftywiranda13: so it's safe to write `eslint --fix`

bhough: @luftywiranda13 Let me double check before I merge this. I was under the same impression, but I think it wasn't respecting that.

bhough:
Damnit @mxstbr! I'm going to put the docs back in. That happens to me all the time on planes. 😬

package.json
bhough: @luftywiranda13 looks like this is actually fine without the path. Thanks for catching that. Not sure what I was seeing earlier lol.

nikgraf:
awesome PR!

PR #186, opened 2017-05-27
fix(color): Prevent floating point issue with alpha channel

timsnadden:
`transparentize(0.9, '#002f6b')` results in `rgba(0,47,107,0.09999999999999998)` and it should be `rgba(0,47,107,0.1)`


bhough:
Thanks for the pr @timsnadden! I will give this a look tonight and see if we can get it out in a release.

nikgraf:
thanks @timsnadden 👌

Issue #176, opened 2017-05-18
toColorString Performance Improvements


bhough:
Performance improvements to color modules by splitting `toColorString` into `hslToColorString` & `rgbToColorString`.

bhough:
Added to V3 Roadmap (#354)

PR #187, opened 2017-05-28
feat(color): toColorString Performance Improvements

sandiiarov:
Split toColorString into hslToColorString & rgbToColorString
Addresses: #176


bhough:
Thanks for the PR @Alex4S. This is a bit item on our v2 list! I've requested a review from @nikgraf since he did a lot of the work on the color side. One note is, while all internal implementation should be moved over to the new specific methods, we should leave `toColorString` in for now and mark it as deprecated. We are working on introducing new code that will warn the user of this (among many other things).

nikgraf:
The PR itself looks good. My original intention was to keep `toColorString` for people doing more complex custom color manipulation. It's quite useful when dealing with color objects.

If we want to deprecate it then the PR is fine. If now we either copy the code to have it in both or refactor to reuse the `rgbToColorString` and `hslToColorString` in `toColorString`.

@bhough thoughts?

bhough:
@sandiiarov if we could get the conflicts resolved on this, we'd like to get this reviewed/merged ASAP. Let me know if you think you won't have time.

bhough:
Conflicts resolved and pulled into #398.

PR #192, opened 2017-06-15
[V2][WIP] feat(Color): Returns most readable color, black or white

LawJolla:
Uses W3C convention to determine if black or white is more readable (greater luminosity difference)
from a given color

The git hooks returned the following error (so I used --no-verify)

```
/polished/docs-theme/index.js:34
        let prefix = ''
        ^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

```

As part of the precommit build, docs/assets/polished.js and docs/docs/index.html were modified.  I hope that's expected behavior.

Thanks for all of your time and hard work!

### DIFF
--- a/src/color/readableColor.js
+++ b/src/color/readableColor.js
@@ -3,38 +3,42 @@
 import parseToRgb from './parseToRgb'
 import curry from '../internalHelpers/_curry'
 
+const w3cConversion = (rgbColor: number): number => {
+  const colorZeroToOne = rgbColor / 255
+  return colorZeroToOne <= 0.03928
+    ? colorZeroToOne / 12.92
+    : ((colorZeroToOne + 0.055) / 1.055) ** 2.4
+}
+
 /**
  * Selects black or white or best contrast depending on the luminosity of the supplied color.
- * Follows W3C specs for readability
+ * Follows W3C specs for readability at https://www.w3.org/TR/WCAG20-TECHS/G18.html
  *
  * @example
  * // Styles as object usage
  * const styles = {
- *   color: mostReadable('#000'),
- *   color: mostReadable('rgb(255,0,0)'),
+ *   color: readableColor('#000'),
+ *   color: readableColor('papayawhip'),
+ *   color: readableColor('rgb(255,0,0)'),
  * }
  *
  * // styled-components usage
  * const div = styled.div`
- *   color: ${mostReadable('#000')};
- *   color: ${mostReadable('rgb(255,0,0)')};
+ *   color: ${readableColor('#000')};
+ *   color: ${readableColor('papayawhip')};
+ *   color: ${readableColor('rgb(255,0,0)')};
  * `
  *
  * // CSS in JS Output
  *
  * element {
  *   color: "#fff";
+ *   color: "#fff";
  *   color: "#000";
  * }
  */
 
 function readableColor(color: string): string {
-  const w3cConversion = (rgbColor: number): number => {
-    const colorZeroToOne = rgbColor / 255
-    return colorZeroToOne <= 0.03928
-      ? colorZeroToOne / 12.92
-      : ((colorZeroToOne + 0.055) / 1.055) ** 2.4
-  }
   const parsedColor = parseToRgb(color)
   const luminosity =
     w3cConversion(parsedColor.red) * 0.2126 +
--- a/src/color/readableColor.js
+++ b/src/color/readableColor.js
@@ -1,0 +1,46 @@
+// @flow
+
+import parseToRgb from './parseToRgb'
+import curry from '../internalHelpers/_curry'
+
+/**
+ * Selects black or white or best contrast depending on the luminosity of the supplied color.
+ * Follows W3C specs for readability
+ *
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   color: mostReadable('#000'),
+ *   color: mostReadable('rgb(255,0,0)'),
+ * }
+ *
+ * // styled-components usage
+ * const div = styled.div`
+ *   color: ${mostReadable('#000')};
+ *   color: ${mostReadable('rgb(255,0,0)')};
+ * `
+ *
+ * // CSS in JS Output
+ *
+ * element {
+ *   color: "#fff";
+ *   color: "#000";
+ * }
+ */
+
+function readableColor(color: string): string {
+  const w3cConversion = (rgbColor: number): number => {
+    const colorZeroToOne = rgbColor / 255
+    return colorZeroToOne <= 0.03928
+      ? colorZeroToOne / 12.92
+      : ((colorZeroToOne + 0.055) / 1.055) ** 2.4
+  }
+  const parsedColor = parseToRgb(color)
+  const luminosity =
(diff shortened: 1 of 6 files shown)


src/color/readableColor.js
```diff
@@ -0,0 +1,46 @@
...
+ * // CSS in JS Output
+ *
+ * element {
+ *   color: "#fff";
+ *   color: "#000";
+ * }
+ */
+
+function readableColor(color: string): string {
+  const w3cConversion = (rgbColor: number): number => {
```
bhough: Let's pull this out into a separate definition so it isn't redefined every time the main method is run.

src/color/readableColor.js
```diff
@@ -0,0 +1,46 @@
...
+      : ((colorZeroToOne + 0.055) / 1.055) ** 2.4
+  }
+  const parsedColor = parseToRgb(color)
+  const luminosity =
+    w3cConversion(parsedColor.red) * 0.2126 +
+    w3cConversion(parsedColor.green) * 0.7152 +
+    w3cConversion(parsedColor.blue) * 0.0722
+  return luminosity > 0.179 ? '#000' : '#fff'
+}
+
```
bhough: Normally I would love to set `const` names for all these magic numbers so we can understand the code a bit better, but in the interest in keeping the size down that may not be best. Is there an external link or other info we can add to the docs for this that explains what all this math is doing?

src/color/readableColor.js
```diff
@@ -0,0 +1,46 @@
...
+import curry from '../internalHelpers/_curry'
+
+/**
+ * Selects black or white or best contrast depending on the luminosity of the supplied color.
+ * Follows W3C specs for readability
+ *
+ * @example
+ * // Styles as object usage
+ * const styles = {
+ *   color: mostReadable('#000'),
```
bhough: Need to update the docs to reflect `readableColor` as the method name.

src/color/test/readableColor.test.js
```diff
@@ -0,0 +1,36 @@
...
+  })
+
+  it('should return white given this gray', () => {
+    expect(readableColor('rgb(118,118,118)')).toMatchSnapshot()
+  })
+
+  it('should return black given this gray', () => {
+    expect(readableColor('rgb(117,118,118)')).toMatchSnapshot()
+  })
+})
```
bhough: Can we add some tests for hsl and named-css colors as well?

Also, what is the expected behavior when passed an rgba or hsla color?

bhough:
@LawJolla Thank you for the PR. Have a few comments/questions.

You shouldn't need to be running that with the no-verify flag. Will need to look into why that is happening for you.

The modifications to index.html and polished.js are expected.

src/color/readableColor.js
LawJolla: Originally I had the function outside of readableColor in the same file, but I noticed that on build, it looked like documentation took the first function name.  Should I place it below the readableColor function and define it as a named function?

src/color/readableColor.js
LawJolla: Absolutely.  https://www.w3.org/TR/WCAG20-TECHS/G18.html

The only catch is that I've taken the liberty of doing some algebra that may not be apparent on the surface of that link.  I'll lay it out and see what you think.

The contrast equation is:  (L1 + 0.05) / (L2 + 0.05), where L1 is the luminance of lightest color and L2 the luminance of the darkest.  We know that the luminance of white is 1 and black is 0, so we can do some substitution... 

(L + 0.05) / (0.0 + 0.05) > (1.0 + 0.05) / (L + 0.05)  

If the left side is greater, use black, or else use white.  That equation reduces to:

if L > 0.179 use #000000 else use #ffffff

The rest of the constant are obvious by looking through the link.

src/color/test/readableColor.test.js
LawJolla: I will add those tests!  I hadn't thought of named css.  Good one.  Hummmm.  I'll do some work on that.

The alpha colors are an issue because this function isn't applicable.  We're trying to get the most readable color given another color, but the answer for rgba/hsla depends on the backing color.  (e.g. rgba(0,0,0,0.1) in front of white should return black, but 0,0,0,0.1 in front of black should return white)

I could add an argument to the function for the backing color, but I don't think that's particularly helpful.  In most cases if you know the backing color, you should use a rgb/hex/hsl value and save the compute time.  For nearly all alpha uses, the backing color will be dynamic.  

Is it breaking with library's ethos if the function returns an error to rgba/hsla?

src/color/readableColor.js
LawJolla: I blame Webstorm and its select all feature for that blunder!  haha  Will fix!

src/color/readableColor.js
bhough: you can move it outside and before like normal. Just move the documentation block so it is just before the function you are documenting.

src/color/readableColor.js
bhough: 👍

src/color/readableColor.js
bhough: Since it is already not 1 to 1 to the original docs, why don't we just compress down the math as much as possible and then link to that page in our docs. I plan on doing a pass across the library for v2 to try and reduce size as much as possible, so this would likely get swept up in that anyway.

src/color/test/__snapshots__/readableColor.test.js.snap
```diff
@@ -0,0 +1,31 @@
...
+
+exports[`readableColor should return black given white, rgb(255,255,255) 1`] = `"#000"`;
+
+exports[`readableColor should return white given black, #000 1`] = `"#fff"`;
+
+exports[`readableColor should return white given black, rgb(0,0,0) 1`] = `"#fff"`;
+
+exports[`readableColor should return white given black, rgba(0,0,0,0.1) 1`] = `"#fff"`;
+
+exports[`readableColor should return white given black, rgba(0,0,0,0.7) 1`] = `"#fff"`;
```
bhough: For now we can leave this as is. As soon as #111 is done, we can validate it and choose to either throw a warning or an error when passed rgba/hla.

src/color/test/__snapshots__/readableColor.test.js.snap
```diff
@@ -0,0 +1,31 @@
...
+
+exports[`readableColor should return white given blue, #0000FF 1`] = `"#fff"`;
+
+exports[`readableColor should return white given blue, hsl(250, 100%, 50%) 1`] = `"#fff"`;
+
+exports[`readableColor should return white given blue, hsla(250, 100%, 50%, 0.2) 1`] = `"#fff"`;
+
+exports[`readableColor should return white given rgb(117,117,117) 1`] = `"#fff"`;
```
bhough: Any edge case colors we need to test for to make sure the math doesn't break? Something near the middle of the luminosity spectrum perhaps?

LawJolla: I have such tests for rgb -- rgb(120,120,120) // black, rgb(117,117,117) // white -- but I'll add the same tests for hsl and hex.

nikgraf:
nice addition @LawJolla

PR #194, opened 2017-06-17
refactor(dependencies): move `semantic-release` entirely to `travis`

luftywiranda13:
`semantic-release` is supposed to be run only in Travis,
so it's better to move it there.

I did the same PR to glamorous
https://github.com/paypal/glamorous/pull/156


.travis.yml
```diff
@@ -7,6 +7,7 @@ node_js:
   - '6'
 before_script:
   - yarn global add codecov
+  - yarn global add semantic-release
```
bhough: Don't we still need to add semantic release here?

luftywiranda13: `semantic-release-cli setup` that we ran for the first time to setup semantic-release is just a one-time task to set `gh_token` and `npm_token` on Travis.
after that, the `semantic-release` script has nothing to do locally in a project, it's always meant to be run on travis.
so we can safely move it to Travis, it'll cut down the dependencies too.
I did this to some of my OSS project and it's working just fine

Issue #178, opened 2017-05-19
Encourage users to import specific modules


kof:
I realize that `import {lighten} from 'polished'` looks smaller, but its  bad idea to promote this style, because of the growing bundle, much better would be to show `import lighten from 'polished/lighten'` in the examples.

bhough:
@kof @mxstbr I don't believe that this syntax will actually work with the way `polished` is built currently. This would require use to build each module as a separate package inside of the main package for that import style to work (unless there's another way I'm not aware of).

The current suggested method works for bundlers that have tree-shaking, but it's debatable how widespread it's use is, so I'm not opposed to this change. However, we should be explicit about what we are asking for in this issue, as it is quite a bit more work than just updating the example.

silverlight513:
@bhough, Would you mind giving an example of how you got this to work? 

I'm using webpack 2 as my bundler and I'm getting an error when I try this method. After looking at my bundle using webpack-bundle-analyzer I'm pretty sure that tree shaking isn't removing the packages I'm not using. Polished is standing at 8.58KB (gzipped) when I'm only using rem

```
Error: Cannot find module 'polished/rem'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._resolveFilename (/Users/jackrimell/widget/node_modules/module-alias/index.js:33:29)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/jackrimell/widget/client/components/SearchForm.js:3:1)
    at Module._compile (module.js:571:32)
    at loader (/Users/jackrimell/widget/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/jackrimell/widget/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:488:32)
```

bhough:
@SilentCicero my comment probably will cause some confusion...

What currently works is `import rem from 'polished/lib/helpers/rem'`. I'll update my original comment to reflect that we still need to update the build if we want the shorthand `polished/rem` to work.

Is that your production bundle your are analyzing? Should see comments like this `/* unused harmony export triangle */ in your dev bundle to know that tree-shaking is working.

silverlight513:
I was looking at my production build. I'll have a look at switching to that import from libs. Thanks :)

bhough:
[...]

PR #205, opened 2017-06-25
perf(modules): Change module build to better facilitate individual imports

bhough:
Addresses comments: #178 

This will support `import <module> from 'polished/utils/<module>'`.

This is non-breaking so folks using `import <module> from 'polished/lib/<type>/<module>'`. However, I would like to remove this support in V2 as that was never the official recommendation.

Of course normal destructured imports will continue to work as before.


bhough:
@kof @nikgraf @mxstbr So this is the simplest approach to get us to what was described in #178. It basically moves the build individual modules out of `/lib` to the root of the package. Thoughts?

bhough:
@mxstbr I thought about that or even just putting them all flat in the root. It's not a pretty solution organizationally, but easier to import.

Thoughts?

nikgraf:
@bhough that's a pity, I was hoping we could avoid direct imports …

bhough:
@nikgraf were you hoping we could avoid them due to tree shaking handling that for us?

mxstbr:
Putting that all flat in the root is super duper annoying for development as you'll have dozens of files just sitting there cluttering your editor, next to having to remember to add all new utils to the `.gitignore` etc.

I'd much prefer just having them in a folder. Maybe `/u/` to be short?

PR #208, opened 2017-07-03
feat(typings): Generate typescript definition files

ForbesLindesay:
This is a duplicate of/alternative to #154.  Instead of hand-writing declarations for typescript, I wrote a tool to auto-generate typescript definitions from flow type definitions.

I've only tested it against the polished codebase, so it will probably need extending over time.

This implementation has the advantage of being slightly more complete (I built proper currying support, and files remain typed even if you require them individually as 'polished/lib/whatever'). It is also guaranteed to be kept up to date as the source code changes.

It does force you to add some extra flow annotations, as it will not infer anything much.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ script:
   - npm test
   - npm run lint
   - npm run flow
+  - npm run build:flow
+  - npm run build:typescript
+  - npm run typescript
 after_success:
   - codecov
   - 'curl -Lo travis_after_all.py https://git.io/travis_after_all'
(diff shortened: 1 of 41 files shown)


mxstbr:
This is awesome! Why are there so many changes to the docs though?

ForbesLindesay:
I have no idea. I didn't intentionally modify them.

bhough:
@ForbesLindesay @mxstbr I think the doc issue is due to having an outdated version of documentation.js in your node_modules. Can you try blowing them out and re-running `yarn` or `npm i` and see if that fixes them please?

rohmanhm:
This is awesome

ForbesLindesay:
@bhough It doesn't seem to make any difference.  I do get the following errors though:

```
Table of contents defined sorting of triangle but no documentation with that namepath was found
Table of contents defined sorting of Ratio but no documentation with that namepath was found
Table of contents defined sorting of TimingFunction but no documentation with that namepath was found
```

bhough:
@ForbesLindesay Thanks for giving it a shot. I'll pull down your PR this week and get it fixed. This has come up before unfortunately 😠

bhough:
Closing in favor of #215 which has the documentation fixes referenced above.

Issue #201, opened 2017-06-20
Add Flow Instructions to README


bhough:
We need to add a section to the README with instructions for using `✨ polished` with `flow`. The first step is to add a section for configuring `flow` to ignore the `✨ polished` source code. 

A similar example can be found in the `styled-components` docs: https://www.styled-components.com/docs/api#flow

bhough:
Addressed by #216

PR #216, opened 2017-07-24
docs(flow): Add .flowignore suggestion to docs

bhough:
Add section to docs on how to handle flow errors using .flowignore

#201


wtgtybhertgeghgtwtg:
Wouldn't it be better to have the flow for `polished` work without configuration?  Most of the errors boil down to publishing tests, using non-standard suppression comments, and some arity issues.

bhough:
@wtgtybhertgeghgtwtg that is the mid-term plan, but right now this is for existing users so they can get around the current error. Work is being done to get first-class support up to snuff.

PR #204, opened 2017-06-23
fix(color): Improve the library size for webpack

iamakulov:
Follow-up after https://twitter.com/b_hough/status/878015824056942592

PR #217, opened 2017-07-24
feat(Color): Returns most readable color, black or white

bhough:
Uses W3C convention to determine if black or white is more readable (greater luminosity difference).

All credit to @LawJolla. This is based on #192 to merge into master.

@nikgraf @mxstbr Decided to not hold off on this until V2. Should mirror the above (already approved) PR.


bhough:
Updating to add tree-shaking friendly exporting per #204

PR #220, opened 2017-07-29
Add Size Limit

ai:
I read the good post about [reducing `✨ polished` size](https://iamakulov.com/notes/polished-webpack/).

Good optimization will also add automatically metrics to prevent regression. So I add [Size Limit](https://github.com/ai/size-limit). This tool will calculate the real cost of `✨ polished` for users.

If any PR will add an unexpected big dependency, Travis CI will warn you about increased size.

PostCSS, MobX and Material-UI use Size Limit too :).

@mxstbr @iamakulov what doy ou think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ script:
   - npm test
   - npm run lint
   - npm run flow
+  - npm run size
 after_success:
   - codecov
   - 'curl -Lo travis_after_all.py https://git.io/travis_after_all'
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "postcommit": "validate-commit-msg",
     "lint": "eslint src",
     "flow": "flow check",
+    "size": "size-limit",
     "docs": "pushstate-server docs",
     "prepublish": "npm run build"
   },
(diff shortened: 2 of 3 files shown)


bhough:
@ai Thank you for the suggestion. Does Size Limit support this feature on a per module basis? We really aren't too concerned with the overall libraries size as we do everything we can to encourage users not to import the entire library (through things like tree-shaking and specific imports). Though it would definitely be useful if we can use it to keep individual modules from getting too big as we make changes to the library.

ai:
@bhough yeap, you can specify any input files. Do I understand correctly that you have >10 modules? In this case I see two problems:

1. Specifying every module will make `package.json` big.
2. A developer could forget to add Size Limit definition during adding new module.

In terms of regression protection, there is no big difference between specifying using bundle or specifying separated limit for every module. Both ways will work the same.

But I know one better solution. You can call `npx size-limit --why` and get per-module statistics with Webpack Bundle Analyzer (real screenshot of your project):

emtbzq8

Is it that per-module behavior that you want?

bhough:
@ai Yes, this library is basically a lodash for CSS-in-JS so most users only use specific parts of the library for their specific needs.

The problem this would ideally be able to solve for us to catch when underlying implementation changes increase the size of associated modules. For example.

- I add several new color modules (knowing this goes above the total library limit wouldn't be very useful as this would be expected).
- As part of these new color modules changes are made to `toColorString` that increase the size of the set of color modules that rely upon it. This may not be obvious to a user adding the module that those changes impact many other modules. That is what we would want to catch.

ai:
Hm. Interesting case. I will think about best solution (maybe add something to Size Limit). Give me few days.

bhough:
@ai Going to close this one, please do open another when you have a solution for our unique situation.

Issue #197, opened 2017-06-17
Flow errors


stevenmusumeche:
- `polished` version: 1.1.3
- `JSS-in_CSS` library and version: styled-components:2.0.1
- Any relevant JS library and version: React 15.5.4

Mixin/Helper/Shortand you were using and how you were using it:

```javascript
import { darken } from 'polished';
```

What you are seeing:

400+ flow errors.

embed/node_modules/polished/lib/color/parseToRgb.js.flow:56
 56:       alpha: parseFloat(`${rgbaMatched[4]}`, 10),
                                                  ^^ unused function argument
   19: declare function parseFloat(string: mixed): number;
                                  ^^^^^^^^^^^^^^^^^^^^^^^ function type expects no more than 1 argument. See lib: /private/tmp/flow/flowlib_3515cb84/core.js:19

embed/node_modules/polished/lib/color/parseToRgb.js.flow:83
 83:       alpha: parseFloat(`${hslaMatched[4]}`, 10),
                                                  ^^ unused function argument
   19: declare function parseFloat(string: mixed): number;
                                  ^^^^^^^^^^^^^^^^^^^^^^^ function type expects no more than 1 argument. See lib: /private/tmp/flow/flowlib_3515cb84/core.js:19

What you expected to see:

Either no flow errors from the library, or documentation (like in styled components repo) on how to configure flow to ignore these errors while still allowing the library to be used with flow.

bhough:
Thank you for filing the issue @stevenmusumeche.

We've created #200, #201, and #202 as action items out of this issue. We'd love help with any of the items if you are interested.

Closing in favor of above mentioned issues.

PR #222, opened 2017-07-30
feat(Flow): Type definition improvements

bhough:
- Updates all modules with explicit return types.
- Fixes a variety of arity issues with types for modules with optional params.
- Remove unnecessary suppressions with latest versions of Flow.
- Moves to standard `//$FlowFixMe` for zero config suppressions.
- Updates typing for `_statefulSelectors`-based modules.
- Updates to Flow .51
- Adds `/lib` to ignore in .flowconfig.

Partially addresses #197


src/color/toColorString.js
```diff
@@ -6,38 +6,33 @@ import hsl from './hsl'
...
 
-const isRgb = (color): boolean =>
-  typeof color === 'object' &&
+const isRgb = (color: Object): boolean =>
   typeof color.red === 'number' &&
   typeof color.green === 'number' &&
   typeof color.blue === 'number' &&
-  // $FlowIgnoreNextLine not sure why this complains
-  typeof color.alpha !== 'number'
+  (typeof color.alpha !== 'number' || typeof color.alpha === 'undefined')
```
wtgtybhertgeghgtwtg:
```
 || typeof color.alpha === 'undefined'
```
Does this just silence the Flow error?

bhough: yes, though I'm still working my way through this file, as it is the sole one left keeping us from having full flow support. It is proving...troublesome.

PR #228, opened 2017-08-11
Use 'break-word' as value in wordWrap example.

blicksky:
"break-all" is not a valid value for overflow-wrap and its alias word-wrap. It's only applicable to the non-standard word-break property. The example in the docs was inconsistent with the logic which accounted for this.

### DIFF
--- a/src/mixins/wordWrap.js
+++ b/src/mixins/wordWrap.js
@@ -6,19 +6,19 @@
  * @example
  * // Styles as object usage
  * const styles = {
- *   ...wordWrap('break-all')
+ *   ...wordWrap('break-word')
  * }
  *
  * // styled-components usage
  * const div = styled.div`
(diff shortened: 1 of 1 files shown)


bhough:
Good catch @blicksky! Thank you for taking the time to submit a PR.

blicksky:
I ran `yarn build` on my branch, and it produced many changes to docs/assets/polished.js and docs/docs/index.html. The changes to polished.js look like formatting changes, such as removing quotes from object keys, and in addition to the expected changes to index.html, URLs are updated to point to my forked repo. Is there something I'm missing about how to prepare and commit the doc changes?

bhough:
@blicksky nope you are doing it right. Looks like there is a bug in the documentation autogeneration from forks. I'll merge this and regenerate from the main repo and file a bug to look at this. Thanks for taking a look.

Issue #232, opened 2017-08-22
Transition helper to write dry transition declarations


emattias:
I often want to give 2 or more properties the same transition but I don't want to give it to all properties.

My thinking is that the function takes an array of css properties and the transition string to apply to all properties. Something like this:

```javascript
export const multiPropertyTranstition = (properties, transitionString) => {
  return properties
    .map((property) => {
      return `${property} ${transitionString}`
    })
    .join(', ')
}

```

Not sure if the transitions settings to be applied to all properties should be just a string to suffix (like in my example above) or separate arguments.. 🤔

bhough:
@emattias definitely think this is something we can/should support. We could actually build this into the existing transition shorthand, by type checking the initial arg and if it is an array apply the next argument to all those properties. API would look something like this:

```javascript
// Multi-Property
transitions([color, background-color], '2.0s ease-in 2s')
// Single Property
transitions('opacity 1.0s ease-in 0s', 'width 2.0s ease-in 2s')
```
Is this something you'd be interested in putting together a PR for?

bhough:
Addressed by 2.0: https://github.com/styled-components/polished/milestone/1

PR #238, opened 2017-08-27
feat(transitions): Same transition to multiple properties

bhough:
Transitions can now apply the same values to multiple properties.

Addresses #232


bhough:
Awaiting tuples support in `tsgen` https://github.com/ForbesLindesay/tsgen/issues/2

cc: @ForbesLindesay

bhough:
Awaiting tuples support in `tsgen` https://github.com/ForbesLindesay/tsgen/issues/2

PR #245, opened 2017-09-04
WIP: Fix mix

### LINKED ISSUE #244: First argument of mix() is not optional
- `polished` version: 1.7.0

Mixin/Helper/Shortand you were using and how you were using it:

```javascript
polished.mix('#fff', '#f00');
```

What you are seeing:

```javascript
polished.js:1480 Uncaught Error: Passed an incorrect argument to a color function, please pass a string representation of a color.
    at parseToRgb (polished.js:1480)
    at mix (polished.js:2213)
    at fn (polished.js:1874)
    at <anonymous>:1:1
```

What you expected to see:

```javascript
"#ff7f7f"
```

The first argument of the `mix()` function (the `weight` argument) is not "optional" however in the docs it's explicit that it should be optional and it's implied from the code.
It is optional in the sense that if you pass `undefined` then the default `0.5` will be used but you can't use it with 2 arguments only. 

A solution could be to either refactor the function to check the number of arguments or to update the docs + remove the question mark from the related code. Also the 3rd option is to just modify the docs to reflect that you need to pass `undefined` which doesn't make much sense to me. Neither of these options should be a breaking change.

morajabi:
I didn't find a way to handle typing for this kind of function correctly. I made this PR so you can review it and give your suggestions. (closes #244)

### DIFF
--- a/src/color/mix.js
+++ b/src/color/mix.js
@@ -4,7 +4,11 @@ import rgba from './rgba'
 import parseToRgb from './parseToRgb'
 import curry from '../internalHelpers/_curry'
 
-type Mix = (weight?: number, color: string, otherColor: string) => string
+type Mix = (
+  weight?: number | string,
+  color: string,
+  otherColor: string,
+) => string
 /**
  * Mixes two colors together by calculating the average of each of the RGB components.
  *
--- a/src/color/mix.js
+++ b/src/color/mix.js
@@ -4,6 +4,7 @@ import rgba from './rgba'
 import parseToRgb from './parseToRgb'
 import curry from '../internalHelpers/_curry'
 
+type Mix = (weight?: number, color: string, otherColor: string) => string
 /**
  * Mixes two colors together by calculating the average of each of the RGB components.
  *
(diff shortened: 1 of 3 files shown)


bhough:
@morajabi how we usually handle this is with a union type:

```javascript
function mix(weight: number | string = 0.5, color: string, otherColor: string): string
```

You can look at the rgba module for an example of this pattern where the module functions differently based not the type of the first argument.

morajabi:
I thought there should be something better than this (as it's a bit confusing because of the name of the variables), but I'll go with this now!

bhough:
@morajabi If you wanted to be more explicit you could do a union of two tuples depending on what you were trying to accomplish. Maybe something like:

```javascript
function mix(...args: [number, string, string] | [string, string] ): string {
```

morajabi:
It seems some typings in https://github.com/styled-components/polished/blob/master/src/internalHelpers/_curry.js are not compatible with this typing (more explicit one).

bhough:
Too many type conflicts to get this working without breaking the api/curry. We've updated the docs to be clear on optional usage (for people that may be using `undefined` for weight).

We will potentially address this in a future release.

Issue #262, opened 2017-10-25
Over 90% of `normalize` disappears when building an app in production mode


artisologic:
# TLDR
I believe #253 has introduced unwanted `/*#__PURE__*/` annotations on certain lines of code, resulting in `UglifyJS` deleting said lines when building in production mode.

If you want to follow the process I went through to get to this, continue reading 👇

---

I seem to have hit a very weird issue that only occurs when I build my app in production mode.
I am using a `create-react-app` generated app, so I've included `react-scripts`'s version in the relevant section below:

- `polished` version: **1.8.1**
- `JSS-in_CSS` library and version: **irrelevant**
- Any relevant JS library and version: **react-scripts 1.0.14**

### Mixin/Helper/Shortand you were using and how you were using it:

```javascript
export default function() {
	return injectGlobal`
		${normalize()};
	`;
}
```

### What you are seeing:

Only a very small subset of the `normalize` rules is injected when the app is built in production mode:

![image](https://user-images.githubusercontent.com/1539897/32024243-b781e11e-b9d3-11e7-9c18-3fb3ac5ba04a.png)

### What you expected to see:

The full version of `normalize` is injected correctly when running the app in development mode:

![image](https://user-images.githubusercontent.com/1539897/32024133-621942b2-b9d3-11e7-8299-b6d62efcf243.png)

### Reproduction:

- Generate an app with `create-react-app`.
- Use the `normalize` mixin (logging the results of the function is enough to demonstrate the issue)
- Run in dev mode using `npm start` = 👍 (logs all normalized rules)
- Build in production mode using `npm run build` = 👎 (logs very few normalized rules)

### What I think the issue is:

- The issue started appearing once we upgraded `polished` from `1.7.0` to `1.8.1`.
- After doing a diff between these 2 tags, I couldn't see anything remotely touching the `normalize` mixin.
- I then attempted to look inside the file our module loader was loading, which for us seems to be: `dist/polished.es.js`.
- I noticed in that file that all the rules I see disappear in production all have a `/*#__PURE__*/` comment before the call to `defineProperty(...)` as you can see here:

![image](https://user-images.githubusercontent.com/1539897/32024738-81a0bf64-b9d5-11e7-866e-8e7ba59b7b56.png)
- I then tested to manually remove some of these, and indeed the rules re-appeared in production.
- Once I knew this seemed to be the issue I went back to the diff between `1.7.0` to `1.8.1` and it seems this PR is to blame: #253.
[...]

PR #263, opened 2017-10-27
perf(tooling): Use loose mode for the babel transpilation

Andarist:
See https://github.com/styled-components/polished/issues/262#issuecomment-339904411

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,14 @@
 {
   "presets": [
-    "env"
+    ["env", {
+      "loose": true
+    }]
   ],
   "plugins": [
     "transform-object-rest-spread",
     "add-module-exports",
     "transform-flow-strip-types",
-    "syntax-trailing-function-commas"
+    "syntax-trailing-function-commas",
+    "annotate-pure-calls"
   ]
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -51,7 +51,7 @@ const plugins = [
   babel({
     babelrc: false,
     presets: [
-      ['env', { 'modules': false }],
+      ['env', { 'loose': true, 'modules': false }],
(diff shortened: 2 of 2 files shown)


Andarist:
Ofc it potentially can introduce some issues, its after all a loose mode and is not that strict about following the spec **exactly**, but it provides the same functionality. Spec must assume many different scenarios, which most of them happen only rarely in the code in the wild.

And imho if your code need those speced behaviours it is poorly/bizarrely written and should be changed or it is too clever and should be changed too 😉

Personally I'm using `loose` mode exclusively in all libraries that I maintain (4~) and in all of my apps - never had any issues with it, not myself and not reported by anyone.

Andarist:
ping @bhough , any further thoughts on this?

bhough:
@Andarist I think for a short term fix, I'm willing to go this route. Ideally, I would like to see the plugin move to one of the solutions you suggested (https://github.com/styled-components/polished/issues/262#issuecomment-339904411) as soon as possible. Is there anything I can do to help with that?

Andarist:
@bhough While ofc this fix has fixed only symptoms, I feel like it should stay permanently, most OSS projects use `loose` mode without any troubles - i.e. [react](https://github.com/facebook/react/blob/master/.babelrc), [redux](https://github.com/reactjs/redux/blob/master/.babelrc) and others.

> Ideally, I would like to see the plugin move to one of the solutions you suggested (#262 (comment)) as soon as possible. Is there anything I can do to help with that?

Thanks for the kind offer! I've already sent "defineProperty calls folding" PRs to babel, but they didn't yet get thorough reviews. If we could think about how `annotate-pure-calls` plugin could be resistant against such cases, that would be cool.

Issue #264, opened 2017-10-27
Shorthand for border properties


klshriharsha:
I use this shorthand for border in CSS a lot.
```css
border: 1px solid #000;
```
It would be really convenient to have a function like this.
```javascript
...border('1px', 'solid', '#000')
```
That outputs something like this.
```javascript
'borderWidth': '1px';
'borderStyle': 'solid';
'borderColor': '#000';
```
And/Or functions for each of the four directions like this.
```javascript
...borderTop('1px', 'solid', '#000')
...borderRight('1px', 'solid', '#000')
...borderBottom('1px', 'solid', '#000')
...borderLeft('1px', 'solid', '#000')
```

bhough:
Per @klshriharsha in #265 

> I use a small tool called Fela (https://github.com/rofrischmann/fela) to write styles. It's configured to extract all the CSS in JS and create a CSS file. It does this by taking sets of common styles across the app and creating a CSS class for each set of common styles. But it doesn't try to do this with shorthand properties in CSS like border: 1px solid #000. So instead, if I use borderStyle, borderWidth and borderColor separately, it allows the tool to check if the same are used in other places as well and if so, create a class for it.

aga5tya:
@bhough This looks like a valid use case, I too use border shorthand at various places and this helps with tools like fela and styletron to yield proper atomic class names. Would love to hear if others have an alternative way of handling this use case, esp with object style libs?

bhough:
Addressed in 2.0: https://github.com/styled-components/polished/milestone/1

PR #265, opened 2017-10-28
Feature/shorthand for border

klshriharsha:
I use this shorthand for border in CSS a lot.
```css
border: 1px solid #000;
```
It would be really convenient to have a function like this.
```javascript
...border('1px', 'solid', '#000')
```
That outputs something like this.
```javascript
'borderWidth': '1px';
'borderStyle': 'solid';
'borderColor': '#000';
```
And/Or functions for each of the four directions like this.
```javascript
...borderTop('1px', 'solid', '#000')
...borderRight('1px', 'solid', '#000')
...borderBottom('1px', 'solid', '#000')
...borderLeft('1px', 'solid', '#000')
```

This PR could help with this issue ( #264 )
If this change can be helpful, I could add it to docs as well.

### DIFF
--- a/src/shorthands/border.js
+++ b/src/shorthands/border.js
@@ -3,7 +3,7 @@ import directionalProperty from '../helpers/directionalProperty'
 const propertyMap = ['borderWidth', 'borderStyle', 'borderColor']
 
 /**
- * Shorthand that accepts up to four values, including null to skip a value, and maps them to their respective directions.
+ * Shorthand that accepts three values, one of each of the properties borderWidth, borderStyle and borderColor.
  * @example
  * // Styles as object usage
  * const styles = {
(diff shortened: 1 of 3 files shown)


bhough:
@klshriharsha Since most CSS-In-JS libraries use standard CSS syntax is there a reason why you can't just do?

```javascript
const div = styled.div`
  border: 2px solid #000;
`
```

klshriharsha:
I use a small tool called Fela (https://github.com/rofrischmann/fela) to write styles. It's configured to extract all the CSS in JS and create a CSS file. It does this by taking sets of common styles across the app and creating a CSS class for each set of common styles. But it doesn't try to do this with shorthand properties in CSS like `border: 1px solid #000`. So instead, if I use `borderStyle`, `borderWidth` and `borderColor` separately, it allows the tool to check if the same are used in other places as well and if so, create a class for it.

bhough:
@klshriharsha We would need to see if there was more demand for this use case before including it in the library. It is a pretty narrow use case for us to add a new module to address. I'm going to add your description to the original issue and mark it as a request to gauge demand.

PR #271, opened 2017-11-07
docs(style)Improve color contrast for sidebar text

sashatran:
Proposing this color change to improve accessibility, by using a color that has higher contrast ratio against the background. 

Before: 
original
After: 
proposed

### DIFF
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -169,7 +169,7 @@ h4 {
 }
 
 a {
-  color: #9a255d;
+  color: #0D3523;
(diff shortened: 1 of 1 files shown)


bhough:
Thanks for this @sashatran!

bhough:
@sashatran I actually missed this on review, but what will need updated is actually this file: https://github.com/styled-components/polished/blob/master/docs-theme/assets/style.css. The one in the `docs` folder gets overwritten on every build when the docs are generated.

Issue #250, opened 2017-09-28
Unable to build polished on Windows


jbarco:
Hi,
I forked the project and build it on a Mac successfully (apart from some errors regarding the document generation). However, if I try to build it on a Windows machine, it crashes at the next command:

`tsgen "lib/**/*.js.flow" --ignore "lib/**/_*.js.flow"`

- OS: **Windows 7**
- NodeJs version: **8.6.0**
- npm version: **5.3.0**

```
tsgen "lib/**/*.js.flow" --ignore "lib/**/_*.js.flow"

C:\polished\node_modules\tsgen\lib\walkers\index.js:26
            throw ctx.getError('Unsupported node type', node);
            ^

Error: Unsupported node type

import adjustHue from '../adjustHue'
  3 | 
  4 | describe('adjustHue', () => {
  5 |   it('should adjustHue of a reduced hex color', () => {
  6 |     expect(adjustHue(20, '#448')).toMatchSnapshot()
  7 |   })
    at Context.getError (C:\polished\node_modules\tsgen\lib\Context.js:63:16)
    at Object.walk [as default] (C:\polished\node_modules\tsgen\lib\walkers\index.js:26:23)
    at C:\polished\node_modules\tsgen\lib\index.js:29:76
    at Array.map (<anonymous>)
    at convertFile (C:\polished\node_modules\tsgen\lib\index.js:29:22)
    at C:\polished\node_modules\tsgen\lib\index.js:42:26
    at Array.forEach (<anonymous>)
    at Object.convertDirectory (C:\polished\node_modules\tsgen\lib\index.js:40:34)
    at Object.<anonymous> (C:\polished\node_modules\tsgen\lib\cli.js:16:4)
    at Module._compile (module.js:624:30)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! polished@ build:typescript: `tsgen "lib/**/*.js.flow" --ignore "lib/**/_*.js.flow"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the polished@ build:typescript script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

bhough:
@ForbesLindesay any ideas on this one?

ForbesLindesay:
Not sure, and I'm afraid I don't have easy access to a windows build to test it on.  If you narrow it down to something in tsgen I'd love a pull request :)

bhough:
[...]

PR #295, opened 2018-02-08
build(typescript): Removes typescript and tsgen

### LINKED ISSUE #276: [Typescript] Mismatch between type definitions and effective module export policy
- `polished` version: 1.9.0
- `JSS-in_CSS` library and version: *not relevant*
- Any relevant JS library and version: Typescript 2.5.3 or 2.6

**Mixin/Helper/Shortand you were using and how you were using it:**

```typescript
import readableColor from 'polished/lib/color/readableColor'

console.info('Readable color function:', readableColor)
```

**What you are seeing**: 

`Readable color function: undefined` printed in the console

**What you expected to see:** 
 
`Readable color function: function...` printed in the console

As per the type definitions: 
```typescript
declare function curriedReadableColor(color: string): string

export default curriedReadableColor;
```
`curriedReadableColor` should be exported as es module `default`.

**Reproduction**: [minimal working example](https://github.com/sveinburne/polished-export-bug)

**Inspection and possible fix**
The bug comes from the transpiled source: 
```javascript
// src: node_modules/polished/lib/color/readableColor.js
// ... line 50
exports.default = curriedReadableColor;
module.exports = exports['default'];
```
The last line overwrite `module.exports.default` by direct assignment of the function. If this is the intended building behavior, you should rewrite typescript definitions with `export = curriedReadableColor` ; you could do the same as [lodash does for function direct exports](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/filter.d.ts).

**Remark** The example given at the beginning will not raise problems with any building solution supporting synthetic default exports such as babel (with `transform-es2015-modules-commonjs`) or fusebox. 

**Typescript compiler configuration file**
(note that `tsc` is directly used and sources are run with nodejs 6.12)
```json
{
  "compilerOptions": {
    "target": "es5",
    "module": "CommonJS",
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": false,
    "lib": ["es6", "es7", "dom"],
    "strict": true
  }
}
```

bhough:
Removes typescript tests and tsgen auto typing

BREAKING CHANGE: Anyone using typescript will lose typings with this release.

fix #276 #250


ForbesLindesay:
What!!!!!!!!!!!!!!!!!!!!!!! I put a fair bit of work into adding tyepscript support for this library, and was happy to help maintain it. Why is it suddenly getting removed without warning?

mxstbr:
:thinking_face:

mxstbr:
I'm curious too

ForbesLindesay:
@mxstbr can we revert this and maybe open issues on tsgen for anything that needs fixing to resolve this?

mxstbr:
Yeah sure, I wonder why @bhough removed that though. Haven't stayed up to date with dev of this pkg so don't have any bg information.

bhough:
@ForbesLindesay @mxstbr I'm removing it because it has caused a variety of issues that have either had an issue filed directly (https://github.com/ForbesLindesay/tsgen/issues/2) or have pinged you on on our issue (https://github.com/styled-components/polished/issues/250). And it appears that it may not even be fully working (https://github.com/styled-components/polished/issues/276).  

It has blocked a variety of modules and changes from going live and I needed to move forward with it. In the past the response has been for us to submit a PR, and right now given I'm the only one of us maintaining polished, I don't have the time to also troubleshoot tsgen in order to get items out.

Right now this is not in the production build, so open to a discussion on how to move forward, but as of now I'm leaving it out so I can get things moving forward towards 2.0.

ForbesLindesay:
#250 is a pain for me to solve as I'll have to set up a windows VM to reproduce, but I can have a go if it's truly breaking.

I will add a config option to resolve #276 - I was using the allow synthetic default export option for my typescript app. If you were using the default transform to ES Modules this would work fine anyway, but you're compiling to CommonJS style.

I'll fix the tuples issue, I didn't know it was a big enough problem to lead to you dropping the library on polished.

bhough:
On #250 it has prevented a few contributors from contributing since they are on Windows. We did a bunch of work to ensure the whole build worked in Windows for them, so this resulted in a regression for them.

On #276 does that mean only one or the other will work for typescript apps depending on how they have theirs configured?

I'm planning on going back through the flow defs as part of 2.0 and provide a lot stricter type safety.  Tuples are a huge way of doing this for us, since we have a lot of modules with multiple api signatures.

Very much appreciate the help sorting these issues out.

ForbesLindesay:
> On #276 does that mean only one or the other will work for typescript apps depending on how they have theirs configured?

No, the current type definitions are incorrect. Unlike flow, typescript differentiates between `export default` and `module.exports =`. There is a special mode that lets you treat them as equivalent, but if we get the library definitions correct, it won't matter which mode the consumer is using.  Since you are doing `module.exports = ` you'll need to put tsgen in the "Common JS Default Export" mode that is I'm about to add.

ForbesLindesay:
tsgen 1.1.0 with the `--common-js-default-export` flag should fix everything apart from the windows support.

ForbesLindesay:
Version 1.1.1 should slightly improve the error message, but I can't think of anything that would be causing this to break.  Is this sufficient for you to re-introduce typescript definitions?

lookapanda:
If I can leave my bits here.

I noticed in my own project that:
`"build:flow": "flow-copy-source -v -i '{**/*.spec.js}' src lib"`
is not ignoring `.spec.js` files as it should be, they are still converted to `.js.flow` files.

If I change it to:
``"build:flow": "flow-copy-source -v -i **/*.spec.js src lib"``
then they are successfully ignored and then the `build:typescript` task will also run through.

That seems to be contrary to what the author of that lib said, but that's also what I noticed.
https://github.com/AgentME/flow-copy-source/issues/2

resir014:
Hey, I'm in favour of bringing TypeScript back. Lack of typings will definitely prevent me from upgrading to 2.0 in my projects. Let me know how I could be of any help.

bhough:
@resir014 I have a branch where I'm looking at @ForbesLindesay's changes: https://github.com/styled-components/polished/tree/readd-tsgen if you want to track progress on keeping it in 2.0.

Issue #302, opened 2018-03-19
RFC: Function to return contrast ratio between two colors -> contrastRatio()


brianespinosa:
Currently I am not seeing a way to return the contrast ratio between two colors. I could technically use `getLuminance()` to return the two values separately, then subtract to get the difference. It seems like it would be helpful to be able to instead just pass two colors and get the contrast ratio returned.

**Use case:**
If I wanted to check a background color to make sure a foreground text color was dark/light enough to be legible, I could darken/lighten the foreground color to ensure legibility.

Basically this would be similar to `readableColor()` except you are not limited to only returning black/white based on the single color passed.

bhough:
Added to V3 Roadmap #354

brianespinosa:
Thanks @bhough 👍

PR #305, opened 2018-04-03
feat(color): Add contrastRatio mixin

Raicuparta:
Addresses #302

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -36,6 +36,7 @@
     },
     "adjustHue",
     "complement",
+    "contrastRatio",
     "darken",
     "desaturate",
     "getLuminance",
--- a/docs/assets/polished.js
+++ b/docs/assets/polished.js
@@ -11,8 +11,6 @@ function capitalizeString(string) {
   return string.charAt(0).toUpperCase() + string.slice(1);
 }
 
-module.exports = exports["default"];
-
 //      
 var positionMap = ['Top', 'Right', 'Bottom', 'Left'];
 
@@ -80,14 +78,11 @@ function directionalProperty(property) {
   return generateStyles(property, valuesWithDefaults);
 }
 
-module.exports = exports['default'];
-
 //      
 
 function endsWith (string, suffix) {
   return string.substr(-suffix.length) === suffix;
 }
-module.exports = exports["default"];
 
 //      
 
@@ -117,8 +112,6 @@ function stripUnit(value) {
   if (isNaN(unitlessValue)) return value;
   return unitlessValue;
 }
-
-module.exports = exports["default"];
 
 //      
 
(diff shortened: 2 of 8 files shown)


bhough:
Thanks for submitting this @Raicuparta. First question I have is how would you see this being used in production? I'm just wondering if we are introducing a module that is great during development but doesn't see production use and if there is a way to possibly add to the api of one of the related modules (like getLuminance) for this purpose.

Raicuparta:
@bhough Making a tool that lets the user pick foreground / background colors. This shortcut would  be useful for warning users if they're picking two colors with not enough contrast between them.

Maybe @brianespinosa can chime in since he also has a need for it?

brianespinosa:
@bhough @Raicuparta Good morning!

Our use case here is theming. I work on an enterprise product that allows a user to provide a brand color that will change that color throughout our product. In the case of buttons, we sometimes need to know if we should use white or black text on top of the button for legibility.

Because that brand color could be anything, we actually need that check to happen in production, not development.

For more clarity, in case it's useful information, we are using Fela for creating all of our styles.

bhough:
Thanks for the use cases! @brianespinosa It sounds like https://polished.js.org/docs/#readablecolor would solve that issue unless I'm not fully understanding the use case.

Raicuparta:
From #302:
> Basically this would be similar to readableColor() except you are not limited to only returning black/white based on the single color passed.

This particular case could be solved by having some way of telling readableColor() to return something other than pure black / white, like some optional parameter.

And now that I think about it, I suppose my use case example maybe falls outside the scope of polished, since warning users about contrast problems isn't really style-related.

bhough:
@Raicuparta Maybe there is something in a PR that modifies `readableColor` to allow the user to pass colors to override the black and white it currently returns.

brianespinosa:
Yes, this would work... and probably makes more sense as an optional parameter than adding a new function that does almost the same thing:  >@Raicuparta Maybe there is something in a PR that modifies readableColor to allow the user to pass colors to override the black and white it currently returns.

brianespinosa:
[Here are some examples from TinyColor that are doing similar things](https://github.com/bgrins/TinyColor#readability). I don't think we necessarily need all of these, but this one use case proposed in #302 would be enough to accomplish all of the Readability type functions referenced in TinyColor.

bhough:
@Raicuparta I still think we can make it work, but might need to mess with the typing for the curry module. We are doing something similar with with `mix` to address #244. I haven't dug too deep into it yet though.

brianespinosa:
@bhough Good morning! I am just trying to follow along here. Is this something being resolved elsewhere, or are you just simply closing the old PR out and leaving #302 open to capture the feature req?

bhough:
@brianespinosa Yes, the approach that this takes is ultimately not the way we want to go, and some work needs to be done on the Flow side to allow us to modify readableColor. So when we tackle this is it will be done as a new PR.

PR #307, opened 2018-04-07
Fix link url to contributors' page

cloudever:
I just had seen that link goes wrong, while was meeting with the lib, fixed with a cap of tea :)


bhough:
Thanks for the PR @cloudever 👍

Issue #229, opened 2017-08-11
Documentation auto-generation breaks links when done from a fork


bhough:
When a contributor builds the docs from a fork, it replaces all the links with links to their fork instead of the main repo. These should continue to point to the main repo. @mxstbr would love any insights here since you did a lot of this work.

bhough:
Addressed in 2.0: https://github.com/styled-components/polished/milestone/1

PR #308, opened 2018-04-08
docs(babel): add plugin

ChristopherBiscardi:
Mention the babel plugin in the docs whereas it didn't exist before.

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -15,6 +15,11 @@
       "markdown": false
     },
     {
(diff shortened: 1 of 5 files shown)


bhough:
@mxstbr This needs rolled back or the docs need rebuilt. All the links point to the fork per #229 that we haven't addressed in 1.X.

mxstbr:
Oh oops, my bad

PR #311, opened 2018-05-06
build(size): add size snapshot plugin for rollup

TrySound:
This plugin helps to track how any change affects bundle size in every
commit.

For `esm` bundles it also provides treeshaked size by both rollup and
webpack, so you can see the proof of one or another tool which makes
your code treeshakable.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,0 +1,21 @@
+{
+  "dist/polished.js": {
+    "bundled": 87207,
+    "minified": 27455,
+    "gzipped": 8808
+  },
+  "dist/polished.es.js": {
+    "bundled": 85897,
+    "minified": 31176,
+    "gzipped": 9263,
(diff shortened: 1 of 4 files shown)


bhough:
Thanks for the Pr @TrySound. We've had similar packages suggested in the past (https://github.com/styled-components/polished/pull/220) and always run into the same discussion. 

Because of how `polished` is used, we aren't really concerned about the overall size of the bundle, since users often just import an individual module either directly or via tree-shaking. We'd definitely consider a version of this that can compare the impact a change has on the size of individual module though.

TrySound:
The idea of this package is show how big user bundle will consume (bundled), how much will load browser when every import is used (gzipped), how much browser will parse when every import is used (minified) and the most important is the ability to treeshake out unused code. And as you can see treeshakability is not perfect here.

PR #314, opened 2018-05-20
feat(triangle): Add ability to use any units instead of only px to triangle mixin

### LINKED ISSUE #313: css triangle optimizations
- `polished` version: latest

What you are seeing:

in the example

```
// Styles as object usage

const styles = {
  ...triangle({ pointingDirection: 'right', width: '100px', height: '100px', foregroundColor: 'red' })
}


// styled-components usage
const div = styled.div`
  ${triangle({ pointingDirection: 'right', width: '100px', height: '100px', foregroundColor: 'red' })}


// CSS as JS Output

div: {
 'borderColor': 'transparent',
 'borderLeftColor': 'red !important',
 'borderStyle': 'solid',
 'borderWidth': '50px 0 50px 100px',
 'height': '0',
 'width': '0',
}
```

What you expected to see:

rather than using !important for the borderLeftColor, I think the long hand border properties would be better here, I think we could achieve this by introducing a new helper that splits css values like `100%` and returns either a array or object that gives you the value and the unit. so you can do the following things:

1. specify a triangle that isn't based on pixels.
2. specify a triangle without the `!important`

if this sounds like something this library wants, I'll create a PR.

if you have any suggestions on names for the unit parser function, let me know. my initial thought is calling it `cssLength` as thats what the unit as a whole is called in the CSS Spec, but it does sound a bit weird as well.

Reproduction:

lifeiscontent:
closes #313

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -15,6 +15,11 @@
       "markdown": false
     },
     {
+      "name": "Babel",
+      "description": "<div class='babel'><code class='command'>npm install --save-dev babel-plugin-polished</code><code class='javascript'>{ plugins: [\"babel-plugin-polished\"] }</code></div>",
+      "markdown": false
+    },
+    {
       "name": "Mixins"
     },
     "clearFix",
--- a/.documentation.json
+++ b/.documentation.json
@@ -15,11 +15,6 @@
       "markdown": false
     },
     {
-      "name": "Babel",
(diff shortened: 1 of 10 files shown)


bhough:
@lifeiscontent I switch the base to `version-2` since I will likely roll this out as part of that release.

You have mentioned in the original issue also addressing the `!important` needed in the current triangle module. Were you going to do that as part of this PR?

lifeiscontent:
@bhough fixed! 👍

bhough:
Thanks @lifeiscontent! I will give it a look this weekend and let you know if we need any other changes.

PR #317, opened 2018-06-06
feat(tracking) : Add helper for converting Illustrator, Photoshop, Indesign tracking values to em

edword:
👋 In the event this is helpful!

Helper to convert tracking to letter-space friendly ems. 

Useful for converting Adobe Illustrator, Indesign and Photoshop tracking values, in which tracking is measured in 1/1000 em.

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -83,6 +83,7 @@
     "modularScale",
     "rem",
     "stripUnit",
+    "tracking",
     {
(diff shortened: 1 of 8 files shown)


bhough:
@edword Thank you for the PR. Right now this isn't something we would want to include in polished given it's relative simplicity for someone to either implement in code or just do the math in their head.

PR #323, opened 2018-06-27
Fix not covered invalid input

tobilen:
After upgrading to flow 0.75, i discovered that there is a case of invalid input that gets through the regex of the hsl matcher but not the one extracting color values from rgb. This PR should prevent this possibility.


bhough:
Thanks @tobilen I will take a look and try to get a release cut tonight.

PR #331, opened 2018-07-25
feat(Flow): Define, use & document Styles type for mixins result values

### LINKED ISSUE #261: Mixin typescript definitions are not compatible with styled-components
- `polished` version: 1.8.1
- `JSS-in_CSS` library and version: styled-components 2.2.1
- Any relevant JS library and version: 
  + React 16.0.0
  + Typescript 2.5.3

#### Mixin/Helper/Shortand you were using and how you were using it:

- I was using `normalize`, but any other mixin that returns an `Object` has the same issue.

```typescript
import { normalize } from 'polished';
import { injectGlobal } from 'styled-components';

injectGlobal`${normalize()}`;
```

#### What you are seeing:

When compiling this code with Typescript, I get:
```
error TS2345: Argument of type 'Object' is not assignable to parameter of type 'SimpleInterpolation'.
  The 'Object' type is assignable to very few other types. Did you mean to use the 'any' type instead?
...
```

#### What you expected to see:

The code should compile with Typescript.

vhfmag:
As mentioned in #261, current usage of Object to type style objects is too permissive to be incompatible with `styled-components` type definition for interpolated values in tagged template literals. Usage of Object is also discouraged by Flow in its documentation:

> However, if you need to opt-out of the type checker, and don’t want to go all the way to any, you can instead use Object. Object is unsafe and should be avoided.
> – https://flow.org/en/docs/types/objects/#toc-object-type

In this PR, a Styles type, akin to the one declared in `styled-components`,  is implemented and used wherever relevant. It was declared as follows:
```typescript
export type Styles = {
  [ruleOrSelector: string]: string | number | Styles,
}
```

Running `yarn flow` succeeds with 0 errors and Typescript type definitions work too. The only issue with this PR that I could find is that the generated documentation for the Styles type doesn't seem quite right and I couldn't think of any way to solve it, as I couldn't find a way to declare string index properties with JSDoc:

image

If there are known fixes for that or further error, I would be glad to include them. I hope this can be helpful :)

Closes #261

### DIFF
--- a/.documentation.json
+++ b/.documentation.json
@@ -88,6 +88,7 @@
     },
     "AnimationProperty",
     "FontFaceConfiguration",
+    "Styles",
     "HslColor",
     "HslaColor",
     "InteractionState",
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ import { compose } from 'ramda' // Replace with any compose() function of your c
 import { lighten, desaturate } from 'polished'
 
 // Create tone() helper
-const tone = compose(lighten(10), desaturate(10))
+const tone = compose(lighten(0.1), desaturate(0.1))
 ```
 
 ### Why not `package-xyz`?
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "eslint": "^4.17.0",
     "eslint-config-airbnb-base": "^12.1.0",
     "eslint-plugin-import": "^2.8.0",
-    "flow-bin": "^0.64.0",
+    "flow-bin": "^0.75.0",
     "flow-copy-source": "^1.2.2",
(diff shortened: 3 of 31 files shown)


bhough:
@vhfmag Thanks for this. I will take a look at this over the weekend and see if we can get a release cut before 2.0.

bhough:
@vhfmag Due to some pretty significant changes coming in v2, it made more sense to merge it as part of that. I just have some docs/typing cleanup to do and we should have a beta out, so hopefully shouldn't be too long.

vhfmag:
Ok! In that case, I'll keep an eye not to miss the beta release. Thanks!

PR #357, opened 2018-08-24
Remove style object example from ellipsis

TrySound:
This mixin does not work anymore with object spread since its output is array.


TrySound:
At least this should work with emotion
```js
<div css={[ellipsis(), { fontSize: 14 }]}></div>
```

bhough:
Closing this as it will no longer be needed due to #361.

PR #277, opened 2017-11-17
[WIP] Calculating with unit-suffixed values

tobilen:
adresses https://github.com/styled-components/polished/issues/270

* Adds a `calc` helper
calc provides an idx-like syntax to perform arithmetics on unit-suffixed css values.
Example:
```js
import { calc, stripUnit } from 'polished';

const  base = '12px';
const double = calc(base, baseAsNumber => baseAsNumber * 2); // returns '24px'

const one = '1px';
const plusOne = calc(base, baseAsNumber => baseAsNumber + stripUnit(one)); // returns '13px'
```
* Adds a `extractUnit` helper
counterpart to `stripUnit`, `extractUnit` takes in a unit-suffixed css value and returns the unit type as string. Empty string if no type could be extracted
Example:
```js
import { extractUnit } from 'polished';

const unit = extractUnit('12px') // returns 'px'
```

Issue #270, opened 2017-11-06
Feature Request: Calculating with unit-suffixed values


tobilen:
If i want to perform arithmetic on values that have a unit suffix, i can do it like so in sass:
```ssss
$base: 12px;
$double: 2 * $base;
```

while there isnt really an easy way to make this work in js, polished could provide helper functions that take care of the necessary regex to strip the unit, perform the calculation and return a the result as a string with the original unit. the main benefit here would be that the unit type is retained, which it wont be if `stripUnit` is used.

Mock usage:
```js
import { multiply, add } from 'polished';

const base = '12px';
const double = multiply(base, 2); // returns '24px'

const one = '1px';
const plusOneplusOne = add(base, one, one); // returns '14px'
```

It could also have an [idx](https://github.com/facebookincubator/idx)-like syntax:
```js
import { calc, stripUnit } from 'polished';

const  base = '12px';
const double = calc(base, baseAsNumber => baseAsNumber * 2); // returns '24px'

const one = '1px';
const plusOne = calc(base, baseAsNumber => baseAsNumber + stripUnit(one)); // returns '13px'
```

caribou-code:
I've seen a few similar issue tickets and this linked PR all closed without resolution. Why is this?

Being able to perform calculations on values with units is one of the most basic features in SCSS and I don't see the point in extending styled components functionality with Polished unless it has this sort of thing.

Issue #223, opened 2017-07-31
helper for multiplying font-sizes with units


sagiavinash:
Feature Request: (want to contribute)

i am converting my sass codebase to styled components and i use these sass helpers to multiply pixel font sizes

> roun­d($­num­ber) Rounds a number to the nearest whole number.
> 
> ceil­($n­umb­er) Rounds a number up to the next whole number.
> 
> floo­r($­num­ber) Rounds a number down to the previous whole number.
> 

currently to achieve this, i did
```
import { stripUnit } from 'polished';

const fontSizeBase = '14px';
const fontSizeLarge = `${Math.ceil(stripUnit(fontSizeBase) * 1.25))}px`;
```

a generic helper can be something like this
```
// polished helpers
const multiplyFontSize = (multiplier, fontSize) => {
  const value = stripUnit(fontSize);
  const unit = fontSize.split(value)[1];

  return `${value * multiplier}${unit}`;
};

const roundFontSize = (criteria, fontSize) => {
  const value = stripUnit(fontSize);
  const unit = fontSize.split(value)[1];

  return `${Math[criteria](value)}${unit}`;
};

// helper usage
import { roundFontSize, multiplyFontSize  } from 'polished';

const fontSizeBase = '14px';
const fontSizeLarge = roundFontSize('ceil', multiplyFontSize(1.25, fontSizeBase));
```
any thoughts on this?

sagiavinash:
@bhough any decision or future course of action. I am new to OSS can you guide me?

bhough:
@sagiavinash sorry for the delayed response. So it seems like the helper that would be useful is a general one that allows us to do math with CSS values generally as you've described for the specific font use case. Other issues like #234 are also trying to get at this problem.

So I think a good first step would be to talk through a proposed api for a more general set of helpers for doing math on CSS units. It may take a few attempts as I think I have a firm grasp on the problem, but less firm idea of how we should solve it.

caribou-code:
Is there no progress with this? I recently switched from SCSS where this was core functionality and fundamental to achieving theming, styling consistency and DRY code throughout a web app. I was very surprised to see Polished doesn't already do this.

As a temporary measure, I've implemented a `calculate.js` util in my project to multiply and divide values with units, but this is far from ideal because I have to import this into so many files where I already have polished.

```
// @flow

export const stripUnit = (value: string): ?number => {
  const match = /[0-9]+\.?[0-9]*/.exec(value);
[...]

PR #358, opened 2018-08-24
Calculating with unit-suffixed values

tobilen:
see https://github.com/styled-components/polished/pull/277

i'm opening this PR again on the version-3 branch, seeing as the feature has been added to the 3.0 roadmap ( see https://github.com/styled-components/polished/issues/223#issuecomment-415640661 )

original description:
adresses https://github.com/styled-components/polished/issues/270

* Adds a `calc` helper
calc provides an idx-like syntax to perform arithmetics on unit-suffixed css values.
Example:
```js
import { calc, stripUnit } from 'polished';

const  base = '12px';
const double = calc(base, baseAsNumber => baseAsNumber * 2); // returns '24px'

const one = '1px';
const plusOne = calc(base, baseAsNumber => baseAsNumber + stripUnit(one)); // returns '13px'
```
* Adds a `extractUnit` helper
counterpart to `stripUnit`, `extractUnit` takes in a unit-suffixed css value and returns the unit type as string. Empty string if no type could be extracted
Example:
```js
import { extractUnit } from 'polished';

const unit = extractUnit('12px') // returns 'px'
```


bhough:
@tobilen Thank you for reopening/updating this. We'll ask for some patience in review, we definitely want to include this type of functionality in V3, but need to land a few other PRs and take a bit of a breather from V2. 

There will definitely be a lot of discussion here to make sure we come up with an API that accomplishes all the use cases that have been raised in various issues, but we very much appreciate the contribution and help with this.

bhough:
After taking into account the various requests around this we ended up going with a different interface. A lot of the more complex things people have asked to do become really hard to parse mentally in this style. We often found ourselves doing a lot of nested `calc` calls or overusing `stripUnit`. 

Ultimately, we went a completely different direction focusing on safely parsing whatever formula the user wanted to pass providing a core set of math functions (that can be extended). Given polished's purpose is to make using CSS-in-JS easier for folks coming from straight CSS, this approach better matches that. One of the biggest category of issues we address is folks struggling combining methods within a style block, so we want to avoid that whenever possible.

We really do appreciate the contribution and apologize that a lot of this convo happened behind the scenes and not on this PR. It's not the best contributor experience and we will continue to do better there.

Closed in favor of #398

Issue #354, opened 2018-08-24
V3 Roadmap


bhough:
- [x] toColorString performance improvements (likely based on #187).
- [x] update readableColor to return a custom color instead of just black or white.
- [x] color modules safely handle the `transparent` keyword.
- [x] linear gradient mixin.
- [X] modularScale accepts rem for its base value.
- [x] Math helper for doing operations on values with units.
- [x] Improved errors/validation in development, and stripped with links in production.
- [x] Mark `getUnitAndValue` for deprecation.
- [x] Update `stripUnit` to optionally return the unit that was removed.
- [x] Deprecate selection mixin.
- [x] Deprecate placeholder mixin.

_final_

TrySound:
`sideEffects` is not used by rollup. Pure annotations are used by uglify.

brunolemos:
Maybe re-add the contrast ratio function? I don't see it on the list.

#206 #302 #305 #394

bhough:
@brunolemos https://github.com/styled-components/polished/blob/version-3/src/color/readableColor.js

brunolemos:
Hm, that's more like a sugar for one specific use case, but it's not really a `getContrastRatio` function.

There seems to be a standard for this, the ratio is a number and there's level AA or AAA.
Check https://contrast-ratio.com/

The chrome element style tool works the same way: 

![image](https://user-images.githubusercontent.com/619186/52411084-a19d3500-2ac1-11e9-928c-1f3f6f5839a5.png)

https://developers.google.com/web/fundamentals/accessibility/accessible-styles#color_and_contrast

https://webaim.org/resources/contrastchecker/

bhough:
Understand the use case @brunolemos. Mostly pointing out that most of the issues/PRs you pointed to were resolved by the modifications to `readableColor`. `getContrast` was not something we have ever had and would be a new feature. Happy to accept it as a feature request.

brunolemos:
Ah ok. Seems to be #394 then.

PR #359, opened 2018-08-24
Add error system

Andarist:
Prod builds went down from 9522 to 8892 bytes (min + gzipped).


package.json
```diff
@@ -99,6 +100,7 @@
     "ramda": "^0.25.0",
     "rollup": "^0.64.1",
     "rollup-plugin-babel": "^3.0.7",
+    "rollup-plugin-node-resolve": "^3.3.0",
```
TrySound: Do you think it's worth to add another plugin just to resolve `index.js`?

bhough:
@Andarist Thank you for kicking this off. I'm taking a bit of a breather from V2 before diving in to V3 work, but Ill aim to get some initial comments in next week.

Also, I updated this to be agains the `version-3` branch since this is currently on the roadmap for that release (#354)

Andarist:
Take your time, this doesnt have to wait till v3 though - what I mean is it shouldnt be a breaking change.

bhough:
Pulled this code into #398. Thanks for the contribution @Andarist!

PR #362, opened 2018-08-28
Upgrade to babel 7

TrySound:


package.json
```diff
@@ -123,5 +123,8 @@
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/polished"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.0.0"
```
bhough: Is there a reason this is in dependencies vs devDependencies?

TrySound: This dep is generated in build step. Take a look at dist files

TrySound:
@bhough Hey, no! This is not how babel runtime works. If user doesn't have babel runtime it will break their build!

TrySound:
no!!!!!!!!

bhough:
@TrySound it is a bot not Max. We are testing something, please give us a few minutes.

TrySound:
Make prerelease please. Don't publish broken package.

PR #385, opened 2018-11-06
media mixin for forming media queries

dlwalsh:
In every project I do, I find I'm writing a utility function such as this one to convert my breakpoints into media queries.

It would be nice to have this utility in polished. I think it would be a useful addition to the library.

I submit this PR as proof of concept; it's really intended as a discussion item.


bhough:
Closing due to light interest from the community.

PR #387, opened 2018-11-15
Fix syntax error

jisaacks:
New to polished, reading the docs. Pretty sure this is a syntax error as is.


bhough:
@jisaacks Thanks for catching this

PR #401, opened 2019-02-11
docs: fix typo in setSaturation function

markoradak:
It says 'The lightness range...' where it should say 'The saturation range...'. 😊


bhough:
Thanks @markoradak and congrats on your first contribution to the library!

PR #403, opened 2019-02-12
fix(flow): undefined in tagged template literal with flow-bin 0.92.1

evenchange4:
## What I did
1. upgrade flow-bin to `v0.92.1`
2. fix `undefined in tagged template literal` type error:


bhough:
Thanks @evenchange4! Good ol' flow...

PR #409, opened 2019-02-28
fix: removed Object.values usage in math

### LINKED ISSUE #407: Polished math function breaks IE11 due to Object.values usage
- `polished` version: 3.0.0
- `JSS-in_CSS` library and version: styled-components@4.1.3
- Any relevant JS library and version: react@16.8.0

Mixin/Helper/Shorthand you were using and how you were using it:

```javascript
const start = polished.em('80px')
const end = polished.em('1px')

const ResponsiveDiv = styled.default.div`
width: ${polished.math(`${start} - ${end}`)}; 
```

What you are seeing:

In the console in an IE11 browser you will see the error:
`object doesn't support property or method 'values'`

Causing the page to break and simple render a white-page

What you expected to see:

Ideally I would hope this functionality would support IE11, and either use babel to transpile the Object.values reference in the dist build or simply re-write the usage to avoid using Object.values

Reproduction:

The following code-pen will show the issue:

https://codepen.io/goldsziggy/pen/qvdQoQ?editors=1010#0

However due to codepen having issues with IE11 as well, the best way to see the issue is to fork the code-pen using Chrome/Firefox/Safari.  Then access the URL using IE11, and change the url portion `pen` to the word `debug` like:
https://codepen.io/goldsziggy/debug/qvdQoQ?editors=1010#0

goldsziggy:
Object.values is not supported by IE11, causing this feature to break in
IE11. Converted Object.values to Object.keys.map to simulate the same
behavior.

This will resolve #407

### DIFF
--- a/docs/assets/polished.js
+++ b/docs/assets/polished.js
@@ -495,8 +495,9 @@
     var values = [];
     var pattern = new RegExp( // Pattern for numbers
     "\\d+(?:\\.\\d+)?|" + // ...and patterns for individual operators/function names
-    // Flow does not properly type Object.values (https://github.com/facebook/flow/issues/2221)
-    Object.values(symbolMap.symbols) // longer symbols should be listed first
+    Object.keys(symbolMap.symbols).map(function (key) {
+      return symbolMap.symbols[key];
+    }) // longer symbols should be listed first
     // $FlowFixMe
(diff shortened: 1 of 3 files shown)


bhough:
Closing in favor of #411. Thanks @goldsziggy

PR #435, opened 2019-05-08
fix(triangle): Triangles can inherit their color

### LINKED ISSUE #430: Triangle enhancements break `inherit` as a border color
- `polished` version: 3.2.0

Mixin/Helper/Shorthand you were using and how you were using it:

```javascript
triangle({
    pointingDirection: "top",
    width: `6px`,
    height: `6px`,
    foregroundColor: 'inherit,
})
```

What you are seeing:

A black box

What you expected to see:

A colored triangle

Reproduction:

 https://codepen.io/anon/pen/vMVqyM

It seems that browsers (tested on Firefox and Chrome) do not support the border color inherit in the form

```css
border-color: transparent inherit transparent transparent;
```
but do if they are declared by side.

I would propose to revert back the triangle logic to make it work for inherit (and make the bundle slightly bigger).

This is very useful when creating tooltip arrows who can inherit the color of the border of the tooltip.

sgenoud:
fix #430

### DIFF
--- a/src/mixins/test/__snapshots__/triangle.test.js.snap
+++ b/src/mixins/test/__snapshots__/triangle.test.js.snap
@@ -2,7 +2,8 @@
 
 exports[`triangle should default to a transparent background when not passed a backgroundColor 1`] = `
 Object {
-  "borderColor": "transparent transparent transparent red",
+  "borderColor": "transparent",
+  "borderLeftColor": "red",
   "borderStyle": "solid",
   "borderWidth": "5px 0 5px 20px",
   "height": "0",
@@ -12,7 +13,8 @@ Object {
 
 exports[`triangle should generate a proper triangle when passed all parameters 1`] = `
 Object {
-  "borderColor": "black black black red",
+  "borderColor": "black",
+  "borderLeftColor": "red",
   "borderStyle": "solid",
   "borderWidth": "5 0 5 20",
   "height": "0",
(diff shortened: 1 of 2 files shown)


src/mixins/test/__snapshots__/triangle.test.js.snap
```diff
@@ -83,7 +91,8 @@ Object {
 
 exports[`triangle should properly render bottomRight pointing arrow with blue foregroundColor, width of 20px and height 20px 1`] = `
 Object {
-  "borderColor": "transparent transparent blue transparent",
+  "borderBottomColor": "blue",
+  "borderColor": "transparent",
```
bhough: Unfortunately, key order matters a lot for this particular use case. `borderColor` will override `borderBottomColor` and make the element transparent. We had to do some magic in the original implementation to account for this: https://github.com/styled-components/polished/blob/8607d641e27d945612e81a3078b1674200d6286e/src/mixins/triangle.js

src/mixins/test/__snapshots__/triangle.test.js.snap
```diff
@@ -113,7 +124,8 @@ Object {
 
 exports[`triangle should properly render top pointing arrow with green foregroundColor, width of 20px and height 20px 1`] = `
 Object {
-  "borderColor": "transparent transparent green transparent",
+  "borderBottomColor": "green",
+  "borderColor": "transparent",
```
bhough: Same concern as above here.

src/mixins/test/__snapshots__/triangle.test.js.snap
sgenoud: Is the object always sorted by key or is it just an artifact of the testing framework?

Did you mean using the border color helper or adding `!important` to borderBottomColor?

PR #436, opened 2019-05-09
feat(color): add color contrast helpers

### LINKED ISSUE #394: [Feature Request] Additional color functions
Hello 👋,

I'd like to discuss the possibility of adding more helper functions for color if they seem like a good fit. These are all heavily inspired by the [Color](https://github.com/Qix-/color) library, as well as [Colorable](https://github.com/jxnblk/colorable) by Brent Jackson.

### getColorContrast

Follows [W3's recommended ratio](http://www.w3.org/TR/WCAG20/#contrast-ratiodef) to calculate color contrast.

```js
function getColorContrast(color1, color2) {
  const luminance1 = getLuminance(color1)
  const luminance2 = getLuminance(color2)
  return parseFloat(
    (luminance1 > luminance2
      ? (luminance1 + 0.05) / (luminance2 + 0.05)
      : (luminance2 + 0.05) / (luminance1 + 0.05)
    ).toFixed(2),
  )
}
```

### getColorContrastScores

Again, following the [contrast scores recommended by W3](https://www.w3.org/TR/WCAG20/#visual-audio-contrast).

```js
function getColorContrastScores(color1, color2) {
  const contrastRatio = getColorContrast(color1, color2)
  return {
    AA: contrastRatio >= 4.5,
    AALarge: contrastRatio >= 3,
    AAA: contrastRatio >= 7,
    AAALarge: contrastRatio >= 4.5,
  }
}
```

### isColorLight

This could be used internally in the `readableColor` function.

```js
function isColorLight(color) {
  return getLuminance(color) > 0.179
}
```

### isColorDark

```js
function isColorDark(color) {
  return !isColorLight(color)
}
```

Lastly, [here is an example](https://codesandbox.io/s/53l9qm1m8x) with them in use.

Please let me know if these sound like good additions and I'll be happy to put a PR together 🙂.

souporserious:
Adds color contrast helpers using calculations from W3C.

closes #394


src/types/color.js
```diff
@@ -3,6 +3,19 @@
 // Note: we define properties with JSdoc since documentation.js doesn't recognize
 // exported types yet. See https://github.com/documentationjs/documentation/issues/680
 
+/**
+ * @property {number}  AA
+ * @property {number}  AALarge
+ * @property {number}  AAA
+ * @property {number}  AAALarge
+ */
+export type ContrastScores = {|
```
bhough: Since this is being used as a return type for `getColorContrastScores` each key should return `boolean` as opposed to `number`. This should fix the failing Flow tests in the build.

src/color/getColorContrast.js
```diff
@@ -0,0 +1,23 @@
...
+import getLuminance from './getLuminance'
+
+/**
+ * Returns the contrast ratio between two colors based on
+ * [W3's recommended equation for calculating contrast](http://www.w3.org/TR/WCAG20/#contrast-ratiodef).
+ *
+ * @example
+ * const contrastRatio = getColorContrast('#444', '#fff');
+ */
+export default function getColorContrast(
```
bhough: Wondering if we should just call this `getContrast` to stay in line with `getLuminance`.

src/color/getColorContrastScores.js
```diff
@@ -0,0 +1,24 @@
...
+import type { ContrastScores } from '../types/color'
+
+/**
+ * Returns the contrast scores between two colors based on the
+ * [contrast scores recommended by W3](https://www.w3.org/TR/WCAG20/#visual-audio-contrast).
+ *
+ * @example
+ * const scores = getColorContrastScores('#444', '#fff');
+ */
+export default function getColorContrastScores(
```
bhough: Wondering if this something like `meetsContrastGuidelines` might be better since we really aren't returning scores as much as checking if they comply with the W3s guidelines for different scenarios.

src/color/test/getColorContrast.test.js
```diff
@@ -0,0 +1,42 @@
...
+    ).toMatchSnapshot()
+  })
+
+  it('should return the color contrast of two hsl colors', () => {
+    expect(
+      getColorContrast('hsl(0, 100%, 50%)', 'hsl(0, 100%, 100%)'),
+    ).toMatchSnapshot()
+  })
+
+  it('should return the color contrast of two named CSS colors', () => {
```
bhough: One case not tested (and worth discussing) is how these two modules should return if passed `transparent` as one (or both) of the colors. `getLuminance` returns 0 in these cases, so may be as simple as just testing if we are ok with that. Obviously an edge case, but we do account for it in most color modules.

src/color/test/getColorContrastScores.test.js
```diff
@@ -0,0 +1,47 @@
...
+    ).toMatchSnapshot()
+  })
+
+  it('should return the color contrast scores of two hsl colors', () => {
+    expect(
+      getColorContrastScores('hsl(0, 100%, 50%)', 'hsl(0, 100%, 100%)'),
+    ).toMatchSnapshot()
+  })
+
+  it('should return the color contrast scores of two named CSS colors', () => {
```
bhough: See above about `transparent` use case.

src/color/getColorContrastScores.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+import getColorContrast from './getColorContrast'
+
+import type { ContrastScores } from '../types/color'
+
+/**
+ * Returns the contrast scores between two colors based on the
```
bhough: Should update the description since it is not returning scores but instead checking the codes compliance with various guideline values.

src/types/color.js
souporserious: Ah, totally! I'll get this fixed up.

src/color/getColorContrastScores.js
souporserious: Yeah, I like that name way better! 😄

src/color/test/getColorContrast.test.js
souporserious: Yeah, I wasn't sure how to approach transparency since I believe it could give a false-positive when comparing something like `black` and `transparent`. I'll add some tests around it and then I'd be happy to discuss it further.

souporserious:
Thanks for the feedback @bhough! I'll get these fixed up this week.

souporserious:
@bhough addressed all of your comments. Hopefully, everything should be good now 🤞. Please let me know if there is anything else you see that needs changing.

bhough:
Thanks @souporserious! I will give this all a look this weekend and hopefully get a release out.

PR #438, opened 2019-05-12
Fix(adjustHue): account for potential negative hue value

### LINKED ISSUE #437: adjustHue bug
I've recently discovered Polished, and just love how elegant it is.
However, I've run into a bug in `adjustHue`, which I reproduced in [this codepen](https://codepen.io/DavidDeprost/pen/BeoMGJ).
I really love how well-organised and accessible the Polished codebase is, so I'll try submit a PR later.

- `polished` version: 3.3.0
- `JSS-in_CSS` library and version: styled-components 4.2.0
- Any relevant JS library and version: React 16.8.6

- Mixin/Helper/Shorthand you were using and how you were using it:

  ```javascript
  adjustHue(-25, 'orange'); // ok
  adjustHue(-25, 'cyan');   // ok
  adjustHue(-25, 'red');    // bug: returns black instead of red hue
  ```

- What you are seeing:

  Due to red having a hue of 0, subtracting anything gives a negative hue value,
  which results in black.

- What you expected to see:

  I expected it to work consistently for all colors, so never returning black instead of the desired hue.
  Hue is a value in [0 to 359], so negative values should cycle through to 359 and remain positive,
  regardless of how much was subtracted.

DavidDeprost:
closes #437

### DIFF
--- a/src/internalHelpers/_hslToRgb.js
+++ b/src/internalHelpers/_hslToRgb.js
@@ -22,9 +22,9 @@ function hslToRgb(
   }
 
   // formular from https://en.wikipedia.org/wiki/HSL_and_HSV
-  const huePrime = hue % 360 / 60
+  const huePrime = (((hue % 360) + 360) % 360) / 60 // Corrected for potential negative hue
   const chroma = (1 - Math.abs(2 * lightness - 1)) * saturation
-  const secondComponent = chroma * (1 - Math.abs(huePrime % 2 - 1))
+  const secondComponent = chroma * (1 - Math.abs((huePrime % 2) - 1))
 
   let red = 0
   let green = 0
--- a/src/internalHelpers/_hslToRgb.js
+++ b/src/internalHelpers/_hslToRgb.js
@@ -21,8 +21,8 @@ function hslToRgb(
     return convert(lightness, lightness, lightness)
   }
 
-  // formular from https://en.wikipedia.org/wiki/HSL_and_HSV
-  const huePrime = (((hue % 360) + 360) % 360) / 60 // Corrected for potential negative hue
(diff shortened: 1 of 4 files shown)


src/internalHelpers/_hslToRgb.js
```diff
@@ -22,9 +22,9 @@ function hslToRgb(
   }
 
   // formular from https://en.wikipedia.org/wiki/HSL_and_HSV
```
bhough: Typo here we should fix as part of this.

src/internalHelpers/_hslToRgb.js
```diff
@@ -22,9 +22,9 @@ function hslToRgb(
...
-  const huePrime = hue % 360 / 60
+  const huePrime = (((hue % 360) + 360) % 360) / 60 // Corrected for potential negative hue
```
bhough: Instead of an inline comment, might be worth putting a note in the docs on how we handle the potential negative hue.

DavidDeprost: I fixed the typo and removed the comment. I'll try looking into the docs later, when I have a bit more time.

DavidDeprost:
Just added a test to `hslToRgb`. I'm not familiar with jest, so I wasn't entirely sure if `adjustHue` requires tests. I noticed `adjustHue` doesn't have a single "negative" test defined. It should be pretty straightforward to copy the existing tests but with a negative argument.
Let me know if you feel this would be useful.

DavidDeprost:
Just added a test to `hslToRgb`. I'm not familiar with jest, so I wasn't entirely sure if `adjustHue` requires tests. I noticed `adjustHue` doesn't have a single "negative" test defined. It should be pretty straightforward to copy the existing tests but with a negative argument.
Let me know if you feel this would be useful.

PR #439, opened 2019-05-18
David deprost fix/adjust hue

bhough:


bhough:
Thanks for this @DavidDeprost. I went ahead and updated the docs so we could get this released.

PR #454, opened 2019-07-16
fix(hideVisually): remove `clip-path` from `hideVisually`

ughitsaaron:
I came across this bug in Chrome recently, where the use of `clip-path` forces the browser to make unnecessary repaints. The conventional fix across several other libraries has been to simply fallback to `clip` while the bug is still unresolved. For instance,
* https://github.com/twbs/bootstrap/pull/25197
* https://github.com/h5bp/html5-boilerplate/pull/2025
* zurb/foundation-sites#10914

Thanks for all your hard work on this library!


docs/assets/polished.js
```diff
@@ -1410,10 +1409,15 @@
...
+    // Note: The use of `clip-path` causes performance degredation on scroll events in Chrome.
+    // For more information, see:
+    // * h5bp/html5-boilerplate#2021
+    // * zurb/foundation-sites#10914
+    // * twbs/bootstrap#24906
     return {
       border: '0',
       clip: 'rect(0 0 0 0)',
-      clipPath: 'inset(50%)',
+      // clipPath: 'inset(50%)',
```
ughitsaaron: I chose to comment this out instead of removing the line for posterity and to make it explicit that `clip-path` is _not_ being used.

src/mixins/hideVisually.js
```diff
@@ -32,10 +31,15 @@ import type { Styles } from '../types/style'
...
+  // Note: The use of `clip-path` causes performance degredation on scroll events in Chrome.
+  // For more information, see:
+  // * h5bp/html5-boilerplate#2021
+  // * zurb/foundation-sites#10914
+  // * twbs/bootstrap#24906
   return {
     border: '0',
     clip: 'rect(0 0 0 0)',
-    clipPath: 'inset(50%)',
+    // clipPath: 'inset(50%)',
```
ughitsaaron: Just to explain the decision here: I chose to comment this out instead of removing the line for posterity and to make it explicit that clip-path is not being used.

PR #459, opened 2019-09-26
feat(math): adds round symbol

joekrill:
This adds the ability to use `round()` in the `math` helper, with tests. 

I'm not totally sure what the process is for adding functionality here, and what math functions are going to be supported or not, so I though I'd just put a PR up for.

### DIFF
--- a/src/math/defaultMathSymbols.js
+++ b/src/math/defaultMathSymbols.js
@@ -49,6 +49,10 @@ function min(...a: Array<number>): number {
   return Math.min(...a)
 }
 
+function round(a: number): number {
+  return Math.round(a)
+}
+
 function comma(...a: Array<number | string>): Array<number | string> {
   return Array.of(...a)
 }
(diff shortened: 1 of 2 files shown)


bhough:
@joekrill Sorry for letting this sit so long. We'd likely want to pull this into a new set of math symbols specifically for rounding (round, floor, ceiling) that folks can optionally pull in via `math`'s already built-in extensibility. 

If you're still interested we'd gladly accept a PR to add the new function map.

joekrill:
@bhough I'm not sure what you mean. This is exactly the same way every other math function is implemented. Is there some additional extensibility I'm not seeing? Is there an example of an existing case where you're already doing something similar?

bhough:
`math` has the ability to take in an object of additional functions via the `additionalSymbols` param (https://github.com/styled-components/polished/blob/master/src/math/math.js#L8). If you create a separate file with rounding related functions, users will be able to optionally pass that to the `math` function to leverage those capabilities without us bloating the `defaultSymbolMap`. This is an approach we are planning on taking with existing and future "presets" like the math function or named colors in the color functions.

PR #460, opened 2019-10-04
feat(hsla): enable decimal values in hsla

kyorkston:
Issue: 
parseToRgb's hslaRegex did not allow for decimals in the Saturation and Lightness values.

```
../node_modules/polished/lib/color/parseToRgb.js:156
  throw new _errors["default"](5);
        ^
Error: Couldn't parse the color string. Please provide the color as a string in hex, rgb, rgba, hsl or hsla notation.
```
Caused by: `"hsla(218, 36.4%, 95.7%, 1.0)"`

Explanation:
In our repo we recently changed to using hsla over hex values.  A designer then realised that the colours we translated to were not exact to those in the design system so we are now using a new conversion site (https://www.ginifab.com/feeds/pms/rgb_to_hsv_hsl.html) and this has given us decimal values in saturation and lightness

More test results from regex101 for HSLA:
Screenshot 2019-10-04 at 11 54 21

I'm guessing this would affect HSL as well, but feedback on whether this is a feature you would welcome is wanted before it can be added to the hslRegex.

HSL regex tests:
Screenshot 2019-10-09 at 14 01 50

If this is not something you want to allow in the library it's understandable, feedback is welcome!


bhough:
@kyorkston We definitely would welcome this enhancement. As you pointed out, we would need to do this for HLA as well. If you can add that to this PR, we'd be happy to review it and get it included.

kyorkston:
@bhough that's great, will have the HSL version in soon!

bhough:
Thanks @kyorkston. We will give it a look this weekend and see if we can't cut a release.

PR #466, opened 2019-12-09
Fix spread operator usage

tobilen:
In its most recent version, flow errors on the position shorthand with the following error:

image

This is correct, since position would be overwritten if it was part of the `directionalProperty` function call (currently typed with an indexer in src/types/style.js:6 ). 

With the proposed change, we always take the position that was passed to the function


tobilen:
i'd really like to reenable flow checks for polished. anything i can do to get this merged?

bhough:
@tobilen Thanks for bumping this, it got lost in the shuffle. I'll work to test and merge this today or tomorrow.

PR #484, opened 2020-03-15
feat(readableColor): add strict mode option to meet WCAG AA

### LINKED ISSUE #479: Enhance `readableColor` to support preferences
We've been using Polished more and more in our current project, and find it particularly useful for accessibility.

**The problem**

`readableColor` will return a color that is not accessible whenever the user-supplied `lightReturnColor`/`darkReturnColor` does not contrast strongly enough with `color`.

This is not an issue if the user does not supply `lightReturnColor`/`darkReturnColor` as black/white are returned.

**Proposal**

To resolve this, I propose that we add `strict` flag as a fourth parameter, that ensures the function always returns an accessible color (as per WCAG AA).

```ts
readableColor(color: string, lightReturnColor: string, darkReturnColor: string, strict: boolean)
```

With this flag, if either`lightReturnColor` or `darkReturnColor` do not meet contrast requirements against a given color, `black` or `white` would be returned instead.

This would not be a breaking change.

**Possible change for next major**

In the next major release of polished, this _could_ become the default setting and `strict` could be removed.

---

If you think this change is acceptable, I'll raise a PR. Feedback and other thoughts welcome.

mrmckeb:
Resolves #479

### DIFF
--- a/src/color/readableColor.js
+++ b/src/color/readableColor.js
@@ -1,14 +1,15 @@
 // @flow
+import getContrast from './getContrast'
 import getLuminance from './getLuminance'
-import meetsContrastGuidelines from './meetsContrastGuidelines'
 
 /**
  * Returns black or white (or optional light and dark return colors) for best
  * contrast depending on the luminosity of the given color.
  *
- * Set `strict` to `true` to ensure that the return color always meets or
- * exceeds WCAG level AA or greater. In this case, black or white are used as
- * fallback colors.
+ * When passing custom return colors, set `strict` to `true` to ensure that the
+ * return color always meets or exceeds WCAG level AA or greater. If this test
+ * fails, the default return color (black or white) is returned in place of the
+ * custom return color.
  *
  * Follows [W3C specs for readability](https://www.w3.org/TR/WCAG20-TECHS/G18.html).
  *
@@ -30,7 +31,6 @@ import meetsContrastGuidelines from './meetsContrastGuidelines'
  * `
  *
  * // CSS in JS Output
- *
  * element {
  *   color: "#fff";
  *   color: "#ff8";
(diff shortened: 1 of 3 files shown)


src/color/readableColor.js
```diff
@@ -1,8 +1,15 @@
...
+import meetsContrastGuidelines from './meetsContrastGuidelines'
 
 /**
- * Returns black or white (or optional light and dark return colors) for best contrast depending on the luminosity of the given color.
+ * Returns black or white (or optional light and dark return colors) for best
+ * contrast depending on the luminosity of the given color.
+ *
+ * Set `strict` to `true` to ensure that the return color always meets or
+ * exceeds WCAG level AA or greater. In this case, black or white are used as
+ * fallback colors.
```
bhough: Let's update this to make it clear that this only applies when passing custom colors.

src/color/readableColor.js
```diff
@@ -26,13 +35,30 @@ import getLuminance from './getLuminance'
...
+const defaultLightReturnColor = '#000'
+const defaultDarkReturnColor = '#fff'
+
 export default function readableColor(
   color: string,
-  lightReturnColor: string = '#000',
-  darkReturnColor: string = '#fff',
+  lightReturnColor: string = defaultLightReturnColor,
+  darkReturnColor: string = defaultDarkReturnColor,
+  strict: boolean = false,
```
bhough: Not something you introduced, but let's notate these as optional `?` for Flow.

src/color/readableColor.js
```diff
@@ -26,13 +35,30 @@ import getLuminance from './getLuminance'
...
+
+  // May return a color that does not meet WCAG AA.
+  // TODO: Make `strict` the default behaviour in the next major release.
+  if (!strict) {
+    return preferredReturnColor
+  }
+
+  if (meetsContrastGuidelines(color, preferredReturnColor).AA) {
+    return preferredReturnColor
+  }
```
bhough: You should be able to combine these two checks into an `||` check to reduce verbosity.

src/color/readableColor.js
```diff
@@ -26,13 +35,30 @@ import getLuminance from './getLuminance'
...
+  const isLightColor = getLuminance(color) > 0.179
+  const preferredReturnColor = isLightColor ? lightReturnColor : darkReturnColor
+
+  // May return a color that does not meet WCAG AA.
+  // TODO: Make `strict` the default behaviour in the next major release.
+  if (!strict) {
+    return preferredReturnColor
+  }
+
+  if (meetsContrastGuidelines(color, preferredReturnColor).AA) {
```
bhough: Pulling in `meetsContrastGuidelines` will include a chain of modules that will increase the size of `readableColor`. We might be able to go down the chain a bit (for example just check for the AA value against `getContrast`) to reduce the size some.

src/color/readableColor.js
mrmckeb: I work with TypeScript, and thought this looked wrong - but lazily assumed there was something flow-related going on. Sorry, updated!

src/color/readableColor.js
mrmckeb: I'll take a look - I was trying to balance verbosity and readability. I'm sure I can clean it up more.

src/color/readableColor.js
mrmckeb: Agreed. I also thought about this, but was also concerned with having the same logic living in multiple places.

However, the AA standard is unlikely to change.

mrmckeb:
Thanks again!

PR #514, opened 2020-05-18
feat(color): add isValidColor

### LINKED ISSUE #473: [Feature Request]: isValidColor function?
- `polished` version: latest
- `JSS-in_CSS` library and version: styled-components latest

Could be a handy function for theme editors especially

```javascript
export function isValidColor(rawColor) {
  try {
    Color(rawColor);
    return true;
  } catch (e) {
    return false;
  }
}
```

Arun4033622:
Closes #473


bhough:
@Arun4033622 Thank you for the PR. We likely would look at this for V4, so switched the base branch for you.

Does this method work outside of named colors and hex? If so, would love to see some tests around the other color methods we support (RGB(A), HSL(A), and soon HCL(A). We would need these to be supported in order to include this module.

Arun4033622:
It's works outside of named colors and hex. I'll update the PR with required tests @bhough. :)

Arun4033622:
@bhough I've added tests for RGB(A) & HSL(A). I'm not sure about testing HCL(A) since it is not a legal CSS color value. Any clue?

src/index.js
```diff
@@ -60,6 +60,7 @@ import shade from './color/shade'
 import tint from './color/tint'
 import toColorString from './color/toColorString'
 import transparentize from './color/transparentize'
+import isValidColor from './color/isValidColor'
```
bhough, line 63: This should be alphabetized.

src/index.js
```diff
@@ -145,6 +146,7 @@ export {
   toColorString,
   transitions,
   transparentize,
+  isValidColor,
```
bhough, line 149: This should be alphabetized.

Arun4033622: Made the changes @bhough!

PR #517, opened 2020-05-31
Add prettierrc

ricokahler:
This is an incredibly small thing but as a VS Code user, it's nice to have a prettierrc file top-level vs the command line args because the editor extension uses the prettier config when formatting.

I also tested to see if the lint-staged prettier command would work with the prettierrc config and it did on my machine.


bhough:
Thanks @ricokahler I'm going to merge this with the v4 branch for now so we can continue to test it as we work.

PR #522, opened 2020-06-27
Replace snapshots with specific expect statements

ghostd:


bhough:
@ghostd Thank you for taking the time to do this! We will give it a review over the weekend.

Issue #441, opened 2019-05-26
[Feature Request] Easing gradients


zslabs:
👋 Hi all! I've been a big fan of polished; especially coming from the Sass/PostCSS world. I'm working on a project with Emotion, and am missing a few common PostCSS plugins I used to use; one of them being https://github.com/larsenwork/postcss-easing-gradients

Any thoughts on adding something like that to polished? Thanks!

JacobCooley:
I was curious about this too!

bhough:
We will be adding this in v4.

tomekbuszewski:
> We will be adding this in v4.

Any news on this? I can't find this documented anywhere.

bhough:
@tomekbuszewski We have not released v4 yet. But we are looking at possibly pulling it forward in v3.5 that will be out sometime in the next 4-6 weeks.

ricokahler:
Hey all, around 2 weeks ago, I was emailing @bhough about potentially taking this feature on. After digging into it I have a few suggestions on the API and implementation but it requires some relatively large changes and I want to pitch it before going all in.

## API

I think the best place to integrate this feature is to add it to the existing `linearGradient` and `radialGradient` functions.

Looking at CSSWG Drafts repo, it looks like the syntax [they're converging on (back in 2017 lol)](https://github.com/w3c/csswg-drafts/issues/1332#issuecomment-299990698), is `<color-stop>, <timing-function>, <color-stop>, (repeat)`. Given that, I think an API that makes sense is to mimic that decision and allow timing functions to be provided in the current `colorStops` arg.

E.g.

```js
linearGradient({
  colorStops: [
    '#00FFFF 0%',
    'ease-in-out',
    'rgba(0, 0, 255, 0) 50%',
    'ease-in-out',
    '#0000FF 95%'
   ],
  toDirection: 'to top right',
});
```

If a timing function isn't provided, we default to linear.

## Implementation

[...]

PR #523, opened 2020-06-29
feat: eased gradients

ricokahler:
This PR is still a work in progress but I wanted to pause here for feedback. For the most part, this PR is as planned from [#441](https://github.com/styled-components/polished/issues/441#issuecomment-643717850) but instead of changing the cubic bezier functions to number array's, I opted to parse them via regex since that's how colors are read anyway.

Currently, this PR just adds a few functions that will eventually power eased gradients. The main one is `getScale` which will eventually be imported in linearGradient and radialGradient to enable eased gradients via the color stop array (with no breaking API change).

`getScale` accepts color stops with timing functions in between them. Refer to the tests for usage.

TODO:

- [ ] getScale needs to be integrated with `linearGradient` and `radialGradient`
- [ ] Errors need to use the `PolishedError` with new error messages added to the docs.
- [ ] One gotcha I ran into is that `mix` doesn't mix transparent and non-transparent colors. This restriction should be removed if we want to ease black with transparent
- [ ] Add flow types (I don't know flow, only typescript)
- [ ] Update docs
- [ ] Fix math. Taking a second look, I think the math is wrong for the cubic bezier curve so the equation needs to be fixed but the the approach is still the same

## Screenshots (for `getScale`)

image

image

Here's one with an eased gradient 👇

image

---

Anyway, at this point I think you can see where I'm going with this, and to be honest, I'm not sure I like these changes. It's a lot of code for a relatively niche feature and I'm wondering if you can think of a more simple way to get this done.

Let me know what you think!

cc @bhough


ricokahler:
I just noticed my branch includes some `isValidColor` stuff. I don't know how that got in there but ignore it in the diff for now.

ricokahler:
So I've sat on this for a while and I just don't like this approach. Since there's no easy way to create a timing function, I would advise to not ship eased gradients in v4.

Issue #281, opened 2018-01-03
Docs folder in npm package with OFL fonts


philholden:
Does the docs folder need to be part of the npm dist?

It introduces an unnecessary licence in the dist:
Open Font License (OFL)

bhough:
@philholden this was something the community requested so they could have offline access to the documentation.

philholden:
Thanks for fast reply I appreciate that.

PR #553, opened 2020-11-20
fix: Add additional files and folders to npmignore

scttcper:
Currently, this module is 9mb on disk. Added a bunch of stuff to .npmignore so that they aren't published.

Can check unpacked size via npm https://www.npmjs.com/package/polished
image

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,6 +10,12 @@
 .gitignore
 .npmignore
 .travis.yml
+.browserlistrc
+.prettierignore
+.prettierrc
+.vscode
+babel.config.js
+CODEOWNERS
 
 # folders
 coverage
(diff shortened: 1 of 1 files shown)


bhough:
Thanks for putting this together @scttcper. We will review, but the docs have been an often requested inclusion for offline access: https://github.com/styled-components/polished/issues/281

scttcper:
@bhough kept docs in there. The biggest issue is the .yarn folder which i assume is some sort of cache. Docs is maybe 2mb

bhough:
> @bhough kept docs in there. The biggest issue is the .yarn folder which i assume is some sort of cache. Docs is maybe 2mb

@scttcper That is a remnant of our work in the background to move to Yarn 2. We are taking the [zero-install approach](https://yarnpkg.com/features/zero-installs) with polished once we swap and that folder is where it stores the cache. However, we can remove it for now.

bhough (approved):

PR #559, opened 2021-01-23
chore(scarf): add scarf package analytics dependency

bhough:
Adds scarf.sh dependency to track library and documentation usage.


Turbo87:
@bhough AFAICT using scarf is not GDPR-compliant because the users did not give their explicit approval to be tracked and opt-out is not enough to be compliant. I would strongly recommend to revert this PR.

aviaviavi:
:wave: Hi @Turbo87 and @bhough author of the scarf-js package here, and I'd like to respond to this claim. The `@scarf/scarf` package and associated data practices have been extensively reviewed by Scarf's legal team. The telemetry provided by this package is GDPR compliant.

We are not actually storing personally identifying information, which is a key factor here. IP address metadata is looked up to understand associated company information, but IP's themselves are not being stored. If PII was being stored and/or exposed to Scarf users, you'd be absolutely correct that opt-out wouldn't be enough to be compliant. The opt-out mechanisms we offer are not GDPR considerations, but rather a consideration for how we think OSS analytics like this should behave. There's more info about this on the README of scarf-js if it's helpful. 

I see that this PR has already been reverted, but just want to chime in on the legality of this telemetry. It is our hope that this lets the `polished` contributors better understand how their work is being used, so they can deliver the best software they can to their users. 😄

PR #567, opened 2021-02-17
remove scarf analytics

mansona:
Hey folks 👋 

I'm just opening up this PR to remove the `@scarf/scarf` analytics package from polished. 

I don't want to get into too much of a philosophical conversation about analytics or ethics but I would say that the package is potentially a GDPR violation (as it does not work on an opt-in basis) and there are some serious problems with the implementation of scarf as it currently doesn't allow you opt-out if you are using yarn https://github.com/scarf-sh/scarf-js/issues/3

Cheers!


bhough (approved):

aviaviavi:
Hi @mansona, author of the Scarf package here. From the description for the PR, I just wanted to clarify a few points because it seems there's some misunderstandings here. I already described this a bit in another thread that I happened to see first (https://github.com/styled-components/polished/pull/559#issuecomment-780791004) but to summarize here:

* For yarn: Scarf disables itself entirely during yarn installs, so no analytics are reported at all. It is a firm development guideline for Scarf that analytics are never reported without informing the user, so the analytics hooks are never triggered in a  yarn installation. I'll update our docs to better communicate that.
* No personally-identifiable information is being stored by Scarf, so the consent requirements of GDPR are not triggered here. Scarf offers various opt-out mechanisms on basic principle and for convenience, but not due to legal requirements.

PR #596, opened 2021-12-23
Ignore docs and .husky from npm package

yhnavein:
This reduces npm package size by `1.9MB` and these ignored files are not necessary

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -23,6 +23,8 @@ docs-theme
 src
 test
 flow-typed
+.husky/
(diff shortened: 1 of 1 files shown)


bhough (changes requested):
Hi @yhnavein, thank you for the PR. We specifically keep the docs in the bundle. This has come up several times and we have fallen on the side of leaving them for off-line access: https://github.com/styled-components/polished/issues/281

PR #616, opened 2022-05-19
Refactor triangle to use calc()

roginfarrer:
`triangle()` currently expects height and width values that are a string value with a unit (e.g. `'12px'`) or a number (`12`). It throws an error otherwise.

This prevents the use of CSS custom variables. However, if we change from using JS arithmetic to CSS `calc()`, we can support CSS variables, in addition to number of unit values, without having to do any unit stripping in strings.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-lint-staged --debug
+yarn lint-staged --debug
--- a/src/mixins/test/triangle.test.js
+++ b/src/mixins/test/triangle.test.js
@@ -15,7 +15,7 @@ describe('triangle', () => {
       borderColor: 'black',
       borderLeftColor: 'red',
       borderStyle: 'solid',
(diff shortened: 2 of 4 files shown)


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-lint-staged --debug
+yarn lint-staged --debug
```
roginfarrer, line 4: I can omit this. But I thought it odd that the repo requires `lint-staged` be installed globally when it's a dependency of the project.

src/mixins/test/triangle.test.js
```diff
@@ -267,7 +267,7 @@ describe('triangle', () => {
     )
   })
 
-  it('should throw an error when height or width is not a unit based value.', () => {
+  it.skip('should throw an error when height or width is not a unit based value.', () => {
```
roginfarrer, line 270: Q for the reviewer, do we want to keep this check, but then also explicitly check for a value starting with `var(--`?

PR #626, opened 2022-10-24
[Performance] Improve tint and shade performance

thadeucity:
Improve the performance of the `tint` and `shade` functions.

It is a tiny performance improvement and lib size reduction, but it does not degrade in any shape or form the quality of the code, so I believe it is a welcome change.

The improvement is achieved by exiting the `parseToRgb` function in the first `IF` that checks for `HEX Colors`, this way not testing or matching the default color unnecessarily.

### DIFF
--- a/src/color/shade.js
+++ b/src/color/shade.js
@@ -27,7 +27,7 @@ import mix from './mix'
 
 function shade(percentage: number | string, color: string): string {
   if (color === 'transparent') return color
-  return mix(parseFloat(percentage), 'rgb(0, 0, 0)', color)
+  return mix(parseFloat(percentage), '#000000', color)
 }
 
 // prettier-ignore
--- a/src/color/tint.js
+++ b/src/color/tint.js
@@ -27,7 +27,7 @@ import mix from './mix'
 
 function tint(percentage: number | string, color: string): string {
   if (color === 'transparent') return color
-  return mix(parseFloat(percentage), 'rgb(255, 255, 255)', color)
+  return mix(parseFloat(percentage), '#FFFFFF', color)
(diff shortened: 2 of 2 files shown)


src/color/shade.js
```diff
@@ -27,7 +27,7 @@ import mix from './mix'
 
 function shade(percentage: number | string, color: string): string {
   if (color === 'transparent') return color
-  return mix(parseFloat(percentage), 'rgb(0, 0, 0)', color)
+  return mix(parseFloat(percentage), '#000000', color)
```
bhough, line 30: @thadeucity Any reason that we shouldn't send this as a shorthand '#000'?

src/color/tint.js
```diff
@@ -27,7 +27,7 @@ import mix from './mix'
 
 function tint(percentage: number | string, color: string): string {
   if (color === 'transparent') return color
-  return mix(parseFloat(percentage), 'rgb(255, 255, 255)', color)
+  return mix(parseFloat(percentage), '#FFFFFF', color)
```
bhough, line 30: @thadeucity Same as with `mix`, should we send the shorthand '#fff'?

bhough (changes requested):
Hi @thadeucity, thank you for taking the time to contribute. A couple small changes, and we should be abe to merge this for inclusion in our next release.

src/color/shade.js
thadeucity: The the shorthand one will have two extra Regex validations and  two extra if validations (because the first check is for the normal 6 digit hex and the second one is for the 8 digit hex.

PR #677, opened 2025-03-13
Fix colorToHex conversion

danielfdsilva:
The number should be divided by 255 instead of multiplying.
This assumes the value of color will be 0-255 (which is the case with the `hslToRgb` fn), and not 0-1.


danielfdsilva:
working as expected. closing

